PR #4, opened 2020-05-07
Hub deploy

qiujian16:


manifests/hub/hub-clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: ClusterRole
+metadata:
+  name: system:open-cluster-management:{{ .HubCoreName }}
+rules:
+# Allow hub to monitor and update status of csr
+- apiGroups: ["certificates.k8s.io"]
+  resources: ["certificatesigningrequests"]
+  verbs: ["create", "get", "list", "watch"]
+- apiGroups: ["certificates.k8s.io"]
+  resources: ["certificatesigningrequests/status"]
```
deads2k: I suspect the operator will end up needing approval power, but I guess you'll find out.

manifests/hub/hub-clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
...
+  verbs: ["update"]
+# Allow hub to get/list/watch/create/delete namespace and service account
+- apiGroups: [""]
+  resources: ["namespaces", "serviceaccounts", "configmaps", "events"]
+  verbs: ["get", "list", "watch", "create", "delete", "update"]
+# Allow hub to manage clusterrole/clusterrolebinding/role/rolebinding
```
deads2k, line 20: you'll eventually need the rbac permission required for expanding powers (granting permissions you don't have), but this can get you started.

manifests/hub/hub-clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
...
+- apiGroups: ["rbac.authorization.k8s.io"]
+  resources: ["clusterroles", "clusterrolebindings", "roles", "rolebindings"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+# Allow hub to manage spokeclusters
+- apiGroups: ["cluster.open-cluster-management.io"]
+  resources: ["spokeclusters"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
deads2k: create here seems a bit odd.  I don't remember an automated creation flow

manifests/hub/hub-clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
...
+- apiGroups: ["cluster.open-cluster-management.io"]
+  resources: ["spokeclusters/status"]
+  verbs: ["update", "patch"]
+- apiGroups: ["register.open-cluster-management.io"]
+  resources: ["spokeclusters/clientcertificates"]
+  verbs: ["renew"]
```
deads2k: why does the hub need renewing power?

manifests/hub/manifestworks.crd.yaml
```diff
@@ -0,0 +1,182 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
```
deads2k: This CRD should be copied into the image from the vendored open-cluster-management/api repo

manifests/hub/spokeclusters.crd.yaml
```diff
@@ -0,0 +1,133 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: spokeclusters.cluster.open-cluster-management.io
```
deads2k: same with this CRD. See this for instance: https://github.com/openshift/cluster-config-operator/blob/master/Dockerfile#L10

manifests/hub/manifestworks.crd.yaml
pmorie: Strong +1 on this.

manifests/hub/hub-clusterrole.yaml
qiujian16: correct, I will add escalate on role/rolebinding. This is not needed

manifests/hub/manifestworks.crd.yaml
qiujian16: I think manifestwork/spokecluster crd is installed only when HubCore API is created on hub cluster? Currently I generate bindata with these crds. I will update the makefile, so it will generate bindata from vendor directory.

For nucleus CRD, agree we should copy into image from vendored api repo

manifests/hub/hub-clusterrole.yaml
qiujian16: fixed

manifests/hub/hub-clusterrole.yaml
qiujian16: add the escalate verb.

manifests/hub/hub-clusterrole.yaml
qiujian16: create removed

Makefile
```diff
@@ -7,10 +7,19 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
 	targets/openshift/deps.mk \
 	targets/openshift/images.mk \
+	targets/openshift/bindata.mk \
 )
 
+$(call add-bindata,hub,./manifests/hub/...,bindata,bindata,./pkg/operators/hub/bindata/bindata.go)
+
+copy-crd:
+	cp ./vendor/github.com/open-cluster-management/api/cluster/v1/*.yaml ./manifests/hub/
```
deads2k, line 16: without a verify this will likely slip

Makefile
```diff
@@ -7,10 +7,19 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
 	targets/openshift/deps.mk \
 	targets/openshift/images.mk \
+	targets/openshift/bindata.mk \
 )
 
+$(call add-bindata,hub,./manifests/hub/...,bindata,bindata,./pkg/operators/hub/bindata/bindata.go)
+
+copy-crd:
```
deads2k, line 15: followup - add a verify

pkg/operators/hub/controller.go
```diff
@@ -2,34 +2,295 @@ package hub
...
+	nucleusInformer nucleusinformer.HubCoreInformer,
 	recorder events.Recorder) factory.Controller {
 	controller := &nucleusHubController{
-		kubeClient:         kubeClient,
-		apiExtensionClient: apiExtensionClient,
+		kubeClient:                     kubeClient,
+		apiExtensionClient:             apiExtensionClient,
+		nucleusClient:                  nucleusClient,
+		nucleusLister:                  nucleusInformer.Lister(),
+		currentHubDeploymentGeneration: 0,
```
deads2k, line 75: this will cause excessive rollouts. I suspect you want the OperatorStatus from openshift/api to track it.

pkg/operators/hub/controller.go
```diff
@@ -2,34 +2,295 @@ package hub
...
+	}
+
+	// Render deployment manifest and apply
+	err = n.applyDeployment(config, controllerContext)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	conditions := &hubCore.Status.Conditions
+	if len(errs) == 0 {
```
deads2k, line 160: followup: use Available, Progressing, Degraded like openshift operators.

You can start with lower level FooAvailable and FooDegraded which make more sense here.

pkg/operators/manager.go
```diff
@@ -25,13 +26,23 @@ func RunNucleusOperator(ctx context.Context, controllerContext *controllercmd.Co
 	if err != nil {
 		return err
 	}
-	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, 5*time.Minute)
```
deads2k, line 28: You will need this eventually.

You probably want to consider a thing like https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/starter.go#L76-L99

deads2k:
some followups, but I think this is good enough to start from

/lgtm
/approve

PR #5, opened 2020-05-12
Add spoke operator

qiujian16:


manifests/spoke/spoke-clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: system:open-cluster-management:{{ .SpokeCoreName }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
deads2k, line 8: Let's start smaller.

Can we start with `admin` across all namespaces?  Then add a comment indicating that we want to develop a controller that could work with permissions below cluster-admin.  the tradeoff is responsivity because list/watch cannot be maintained over too many namespaces.

manifests/spoke/spoke-work-deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+      serviceAccountName: {{ .SpokeCoreName }}-sa
+      containers:
+      - name: spoke-agent
+        image: {{ .WorkImage }}
+        imagePullPolicy: IfNotPresent
+        args:
+          - "/work"
+          - "agent"
+          - "--cluster-name={{ .ClusterName }}"
+          - "--kubeconfig=/spoke/hub-kubeconfig"
```
deads2k, line 27: I would expect a hub-kubeconfig flag. `--kubeconfig` is normally for self

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+	"github.com/open-cluster-management/nucleus/pkg/operators/spoke/bindata"
 )
 
-type nucleusAgentController struct {
-	kubeClient kubernetes.Interface
+const (
+	nucleusSpokeFinalizer        = "nucleus.open-cluster-management.io/spoke-core-cleanup"
+	bootstrapHubKubeConfigSecret = "bootstrap-hub-kubeconfig"
+	hubKubeConfigSecret          = "hub-kubeconfig"
+	nucluesSpokeCoreNamespace    = "open-cluster-management"
```
deads2k, line 39: `open-cluster-management-core`

pmorie: typo: nucleus

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+type nucleusSpokeController struct {
+	nucleusClient          nucleusv1client.SpokeCoreInterface
+	nucleusLister          nucleuslister.SpokeCoreLister
+	kubeClient             kubernetes.Interface
+	registrationGeneration int64
+	workGeneration         int64
+}
+
+func init() {
+	utilruntime.Must(api.InstallKube(genericScheme))
```
pmorie, line 58: Do we need to install the schemes for all of these APIs here?

pmorie: (I wouldn't want to take a dep on openshift API just for this method, especially if we only need it to install schemes)

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+	)
+	errs := []error{}
+	for _, result := range resourceResults {
+		if result.Error != nil {
+			errs = append(errs, fmt.Errorf("%q (%T): %v", result.File, result.Type, result.Error))
+		}
+	}
+
+	if len(errs) > 0 {
+		appleErros := operatorhelpers.NewMultiLineAggregate(errs)
```
deads2k, line 193: applyErrors?

manifests/spoke/spoke-work-deployment.yaml
pmorie: It looks to me like this will change to `--hub-kubeconfig` in https://github.com/open-cluster-management/work/pull/10

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+
+func readClusterNameFromSecret(secret *corev1.Secret) (string, error) {
+	if secret.Data["cluster-name"] == nil {
+		return "", fmt.Errorf("Unable to find cluster name in secret")
+	}
+
+	return string(secret.Data["cluster-name"]), nil
+}
+
+func readKubuConfigFromSecret(secret *corev1.Secret, config spokeConfig) (string, error) {
```
pmorie, line 390: typo: kubuconfig

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+
+	// Check if bootstrap secret exists
+	_, err = n.kubeClient.CoreV1().Secrets(config.SpokeCoreNamespace).Get(
+		ctx, config.BootStrapKubeConfigSecret, metav1.GetOptions{})
+	if err != nil {
+		helpers.SetNucleusCondition(&spokeCore.Status.Conditions, nucleusapiv1.StatusCondition{
+			Type:    spokeCoreApplied,
+			Status:  metav1.ConditionFalse,
+			Reason:  "SpokeCoreApplyFailed",
+			Message: fmt.Sprintf("Failed to get bootstracp secret %s/%s", config.SpokeCoreNamespace, config.BootStrapKubeConfigSecret),
```
pmorie, line 167: type: bootstrap

pkg/operators/spoke/controller.go
pmorie, line 193: typo: applyErrors

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+		return err
+	}
+
+	// Deploy the static resources
+	// Apply static files
+	resourceResults := resourceapply.ApplyDirectly(
+		resourceapply.NewKubeClientHolder(n.kubeClient),
+		controllerContext.Recorder(),
+		func(name string) ([]byte, error) {
+			return assets.MustCreateAssetFromTemplate(name, bindata.MustAsset(filepath.Join("", name)), config).Data, nil
```
pmorie, line 180: You need to pass a context here for the template to fill from, correct?

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+		helpers.UpdateNucleusSpokeStatus(
+			ctx, n.nucleusClient, spokeCoreName, helpers.UpdateNucleusSpokeConditionFn(spokeCore.Status.Conditions...))
+		return appleErros
+	}
+
+	// Create hub config secret
+	hubSecret, err := n.kubeClient.CoreV1().Secrets(config.SpokeCoreNamespace).Get(
+		ctx, hubKubeConfigSecret, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		// Craete an empty secret with placeholder
```
pmorie, line 209: create

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+			Reason:  "SpokeCoreApplyFailed",
+			Message: appleErros.Error(),
+		})
+		helpers.UpdateNucleusSpokeStatus(
+			ctx, n.nucleusClient, spokeCoreName, helpers.UpdateNucleusSpokeConditionFn(spokeCore.Status.Conditions...))
+		return appleErros
+	}
+
+	// Create hub config secret
+	hubSecret, err := n.kubeClient.CoreV1().Secrets(config.SpokeCoreNamespace).Get(
```
deads2k, line 206: I don't think I see how you can create this secret.  Where can you get it from?  If the user provides it to you, couldn't the first cut here be the user writing it into the target namespace directly?

deads2k:
I'll merge this to reduce your future diff, but i'd like to see these addressed. I may try to find time this week to make the operators look slightly more openshift-like, which I think will ease things later on.

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+			return err
+		}
+	}
+
+	if err != nil {
+		helpers.SetNucleusCondition(&spokeCore.Status.Conditions, nucleusapiv1.StatusCondition{
+			Type:    spokeCoreApplied,
+			Status:  metav1.ConditionFalse,
+			Reason:  "SpokeCoreApplyFailed",
+			Message: fmt.Sprintf("Failed to get hub kubeconfig secret with error %v", err),
```
pmorie, line 228: `failed to create` seems more appropriate than `failed to get`

pkg/operators/spoke/controller.go
```diff
@@ -2,28 +2,395 @@ package spoke
...
+		})
+		helpers.UpdateNucleusSpokeStatus(
+			ctx, n.nucleusClient, spokeCoreName, helpers.UpdateNucleusSpokeConditionFn(spokeCore.Status.Conditions...))
+		return err
+	}
+	n.registrationGeneration = generation
+
+	// If cluster name is empty, read cluster name from hub config secret
+	if config.ClusterName == "" {
+		clusterName := hubSecret.Data["cluster-name"]
```
pmorie, line 253: I'm asking myself as I read this whether we should formalize an API for what's in this secret.

pkg/operators/spoke/controller.go
qiujian16: I think the bootstrap secret is provided by user, this secret is updated by the registration agent to fill the kubeconfig with cert/key.  Initially it will be empty, and registration agent will update that when csr is approved

manifests/spoke/spoke-clusterrolebinding.yaml
qiujian16: I think `admin` will not be able to create `customresourcedefinitions`. That might be needed for manifestwork. I will add an additional clusterrole for crd together with admin.

PR #6, opened 2020-05-12
Update crd and add verify

qiujian16:
Follow up of pr #4


Makefile
```diff
@@ -18,6 +21,12 @@ copy-crd:
 
 update-all: copy-crd update
 
+verify-crds:
+	$(deps_diff) {vendor/github.com/open-cluster-management/api/cluster/v1,manifests/hub}/0000_00_clusters.open-cluster-management.io_spokeclusters.crd.yaml || ( echo 'crd content is incorrect' && false )
```
deads2k: we need to check all the yamls, right?

pkg/operators/hub/controller.go
```diff
@@ -93,6 +97,10 @@ func (n *nucleusHubController) sync(ctx context.Context, controllerContext facto
 	klog.V(4).Infof("Reconciling HubCore %q", hubCoreName)
 
 	hubCore, err := n.nucleusLister.Get(hubCoreName)
+	if errors.IsNotFound(err) {
+		// HubCoree not found, could have been deleted, do nothing.
```
deads2k: HubCore

Makefile
qiujian16: yes, changes to use script and check all yamls on the dir

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.13
 
 require (
 	github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
-	github.com/open-cluster-management/api v0.0.0-20200506150956-355c9d6ef16b
+	github.com/open-cluster-management/api v0.0.0-20200512021938-5f0d5c12ea67
```
pmorie, line 7: I wonder if we should start tagging releases soon, I have an allergic reaction to go mod pseudoversions

PR #9, opened 2020-05-13
Use operartor-sdk to generate csv

qiujian16:
/hold

need to test when image is ready.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,29 @@
 all: build
 .PHONY: all
 
-IMAGE_REGISTRY?=quay.io
-
 # Include the library makefile
 include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
 	targets/openshift/deps.mk \
 	targets/openshift/images.mk \
 	targets/openshift/bindata.mk \
+	lib/tmp.mk\
 )
+
+IMAGE_REGISTRY?=quay.io/open-cluster-management
+OPERATOR_SDK?=$(PERMANENT_TMP_GOPATH)/bin/operator-sdk
+OPERATOR_SDK_VERSION?=v0.17.0
+operatorsdk_gen_dir:=$(dir $(OPERATOR_SDK))
+KUBECTL?=kubectl
+KUBECONFIG ?= ./.kubeconfig
+OLM_NAMESPACE?=olm
+
+OPERATOR_SDK_ARCHOS:=x86_64-linux-gnu
+ifeq ($(GOHOSTOS),darwin)
+	ifeq ($(GOHOSTARCH),amd64)
+		OPERATOR_SDK_ARCHOS:=x86_64-apple-darwin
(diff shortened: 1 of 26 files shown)


joeg-pro (commented):
@qiujian16 Pper your request for me to take a look:  I think this looks good.  

With this in `nucleus`, we can change the building of the Hub operator bundle to (a) git clone `nucleus` in the `release-2.0` branch, and (b) then use what is found in `deploy/olm-catalog/nucleus` as intput to be merged into the Hub operator bundle.  Our hub build will **not** run any `make` on nucleus` so it will be up to you to refresh the olm catalog stuff and commit the result to master/release-2.0 as changes are needed.   Also, since we will be using the olm-catalog manifest sources from the repo, we do not need  an actual operator bundle docker image to be built for this..

I note that the CSV being generated is skeletal, in that it has no `alm-examples` or `description` etc.  That should not be a problem as we do not need to pick up those values into the merged bundle.  But if you go to publish this as  a community operator independently of the hub, you will want to ge tthe CSV to be more complete so your OLM catalog entry is good for consumers.

When you are moving to publish this as a community operator, please sync with me/install squad before doing that as we will then want to change the approach being used in the Hub bundle generation to correspond.

Makefile
```diff
@@ -2,36 +2,98 @@
...
+update-csv: ensure-operator-sdk
+	$(OPERATOR_SDK) generate csv --crd-dir=deploy/nucleus-hub/crds --deploy-dir=deploy/nucleus-hub --output-dir=deploy/nucleus-hub/olm-catalog/nucleus-hub --operator-name=nucleus-hub --csv-version=0.1.0
+	$(OPERATOR_SDK) generate csv --crd-dir=deploy/nucleus-spoke/crds --deploy-dir=deploy/nucleus-spoke --output-dir=deploy/nucleus-spoke/olm-catalog/nucleus-spoke --operator-name=nucleus-spoke --csv-version=0.1.0
+
+deploy: install-olm deploy-hub deploy-spoke
+
+clean-deploy: clean-spoke clean-hub
+
+install-olm: ensure-operator-sdk
+	$(OPERATOR_SDK) olm status --olm-namespace $(OLM_NAMESPACE) ; if [ $$? -ne 0 ] ; then $(OPERATOR_SDK) olm install --version 0.14.1 ; fi
```
pmorie, line 51: They restored the sha that was preventing 0.15+ from working correctly with this command, so we can probably just install the latest version now.

Makefile
```diff
@@ -2,36 +2,98 @@
...
+	$(KUBECTL) get ns open-cluster-management ; if [ $$? -ne 0 ] ; then $(KUBECTL) create ns open-cluster-management ; fi
+
+deploy-hub:
```
pmorie, line 54: we probably want this to depend on `install-olm` target

pmorie (commented):
I had success with this locally, great start

PR #15, opened 2020-05-22
fix some deploy problems

skeeey:


deploy/nucleus-hub/operator.yaml
```diff
@@ -2,7 +2,7 @@ kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: nucleus-hub
-  namespace: open-cluster-management-core
+  namespace: open-cluster-management
```
deads2k, line 5: I think we just tweaked this open-cluster-management-hub in the api repo here https://github.com/open-cluster-management/api/pull/25

skeeey: This namespace is for nucleus operator, it is not for hub core components, the nucleus operator and hub core have a different namespace, the  hub core namespace is open-cluster-management-hub

PR #16, opened 2020-05-22
add webhook

qiujian16:
add registration webhook
add integration test case for hub operator


pkg/helpers/helpers.go
```diff
@@ -193,3 +231,102 @@ func CleanUpStaticObject(
 	}
 	return err
 }
+
+func ApplyValidatingWebhookConfiguration(
```
deads2k, line 235: open a PR to library-go for this

pkg/helpers/helpers.go
```diff
@@ -193,3 +231,102 @@ func CleanUpStaticObject(
...
+	}
+	if !*modified {
+		return existing, false, nil
+	}
+
+	actual, err := client.ValidatingWebhookConfigurations().Update(context.TODO(), existingCopy, metav1.UpdateOptions{})
+	return actual, true, err
+}
+
+func ApplyDeployment(
```
deads2k, line 262: interesting.  Add godoc.  I'd like to stay in sync with library-go

deads2k:
I think cleanup probably wants a dynamic client and I think we want some bits in library-go or at least  understand why they aren't.  I may bring in some of the openshift cert management, but this looks fine to start

/lgtm
/approve

PR #17, opened 2020-05-22
Test presence of OLM by CSV resource instead of op-sdk

pmorie:


pmorie:
Green-buttoning because I've tested this against kind and openshift locally and want to get CI unblocked

PR #19, opened 2020-05-28
Minor tweaks as I used it.

deads2k:
1. fixes message typo
2. creates namespace if the namespace isn't present
3. fixes a missing condition. conditions should always be set before return
4. tightens expression of condition setting to make it slightly easier to read
5. creates all deployments and then checks to see if those deployments are working. This should be in a separate controller eventually.

### DIFF
--- a/pkg/operators/spoke/controller.go
+++ b/pkg/operators/spoke/controller.go
@@ -157,7 +157,23 @@ func (n *nucleusSpokeController) sync(ctx context.Context, controllerContext fac
 	// Start deploy spoke core components
 	// Check if namespace exists
 	_, err = n.kubeClient.CoreV1().Namespaces().Get(ctx, config.SpokeCoreNamespace, metav1.GetOptions{})
-	if err != nil {
+	switch {
+	case errors.IsNotFound(err):
+		_, createErr := n.kubeClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{Name: config.SpokeCoreNamespace},
+		}, metav1.CreateOptions{})
+		if createErr != nil {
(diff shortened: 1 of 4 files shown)


deads2k:
/hold

this should cause some tests to fail.  If it doesn't, I'll have to add those tests.

pmorie:
This LGTM, @qiujian16 can tag when he is ready.

/retest

pkg/operators/spoke/controller.go
```diff
@@ -228,86 +229,84 @@ func (n *nucleusSpokeController) sync(ctx context.Context, controllerContext fac
...
-			})
-			helpers.UpdateNucleusSpokeStatus(
-				ctx, n.nucleusClient, spokeCoreName, helpers.UpdateNucleusSpokeConditionFn(spokeCore.Status.Conditions...))
+			helpers.UpdateNucleusSpokeStatus(ctx, n.nucleusClient, spokeCoreName, helpers.UpdateNucleusSpokeConditionFn(nucleusapiv1.StatusCondition{
+				Type: spokeRegistrationDegraded, Status: metav1.ConditionTrue, Reason: "ClusterNameMissing",
+				Message: fmt.Sprintf("Failed to get cluster name from `kubectl get secret -n %q %q -ojsonpath='{.data.cluster-name}`.  This is set by the spoke registration deployment.", hubSecret.Namespace, hubSecret.Name),
+			}))
 			return fmt.Errorf("Failed to get cluster name")
 		}
 		config.ClusterName = string(clusterName)
```
qiujian16, line 292: I think we should get cluster name from secret before deploying work agent. If cluster name is not specified in spokecore CR, we will not be able to update the deployment.

qiujian16:
it is neat!

Only need to handle the case of cluster name is not set in spokecore. I will followup on that.

/lgtm

PR #20, opened 2020-05-28
Add image munging for deployment targets

pmorie:
Adds ability to munge the CSVs and operator CRs installed by:

- Setting `IMAGE_REGISTRY` env to set the registry to use
- Optionally overriding entire image by setting `WORK_IMAGE` or `REGISTRATION_IMAGE`


pmorie:
@qiujian16 lgtm is on you

PR #26, opened 2020-06-04
Add deploy instruction on readme

qiujian16:


Makefile
```diff
@@ -92,8 +94,8 @@ cluster-ip:
 
 bootstrap-secret: cluster-ip
 	$(KUBECTL) get ns open-cluster-management-agent ; if [ $$? -ne 0 ] ; then $(KUBECTL) create ns open-cluster-management-agent ; fi
-	cp $(KUBECONFIG) dev-kubeconfig
-	$(KUBECTL) config set clusters.kind-kind.server https://$(CLUSTER_IP) --kubeconfig dev-kubeconfig
+	cp $(HUB_KUBECONFIG) dev-kubeconfig
```
pmorie: i'd expect that kubeconfig and hub_kubeconfig to be the same thing, what am I missing?

qiujian16: i think if we are going to deploy klusterlet on another cluster (not on hub cluster), we have to use another environment var. `KUBECONFIG` will be set as the kubeconfig for spoke cluster, and `HUB_KUBECONFIG` will be set as the kubeconfig for hub cluster

pmorie: I think i would rather have KUBECONFIG always be the hub and have a KLUSTERLET_KUBECONFIG for the spoke, wdyt about that?

qiujian16: that sounds reasonable

mdelder: Any reason not to plan on a single `KUBECONFIG` with multiple contexts?

qiujian16: yes, that is even better. Change to KLUSTERLET_KUBECONFIG_CONTEXT

PR #28, opened 2020-06-08
Fix issue  when externalserver is not set

qiujian16:
1. fixes https://github.com/open-cluster-management/backlog/issues/2599
2. Also fix the issue when clusterName is not set.
3. Add klusterlet integration test

### DIFF
--- a/test/integration/klusterlet_test.go
+++ b/test/integration/klusterlet_test.go
@@ -1,0 +1,225 @@
+package integration
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	"github.com/openshift/library-go/pkg/controller/controllercmd"
+
+	operatorapiv1 "github.com/open-cluster-management/api/operator/v1"
+	"github.com/open-cluster-management/registration-operator/pkg/operators"
+	"github.com/open-cluster-management/registration-operator/test/integration/util"
+)
+
+func startKlusterletOperator(ctx context.Context) {
+	err := operators.RunKlusterletOperator(ctx, &controllercmd.ControllerContext{
+		KubeConfig:    restConfig,
+		EventRecorder: util.NewIntegrationTestEventRecorder("integration"),
+	})
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+}
+
+var _ = ginkgo.Describe("Klusterlet", func() {
+	var cancel context.CancelFunc
+	var klusterlet *operatorapiv1.Klusterlet
+	var klusterletNamespace string
+	var registrationRoleName string
+	var registrationDeploymentName string
+	var registrationSAName string
+	var workRoleName string
+	var workDeploymentName string
+	var workSAName string
+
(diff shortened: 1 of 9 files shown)


manifests/klusterlet/klusterlet-work-deployment.yaml
```diff
@@ -20,10 +20,21 @@ spec:
...
+            valueFrom:
+              secretKeyRef:
+                name: {{ .HubKubeConfigSecret }}
+                key: cluster-name
+          {{end}}
         args:
           - "/work"
           - "agent"
-          - "--spoke-cluster-name={{ .ClusterName }}"
+          - "--spoke-cluster-name=$(CLUSTER_NAME)"
```
deads2k: doing it like this means that when the cluster-name changes, you will not automatically trigger a deployment, right?

test/integration/util/util.go
```diff
@@ -43,3 +46,20 @@ func (r *IntegrationTestEventRecorder) Warning(reason, message string) {
 func (r *IntegrationTestEventRecorder) Warningf(reason, messageFmt string, args ...interface{}) {
 	r.Warning(reason, fmt.Sprintf(messageFmt, args...))
 }
+
+func HaveCondition(conditions []operatorapiv1.StatusCondition, expectedType string, expectedStatus metav1.ConditionStatus) bool {
```
deads2k: HasCondition

manifests/klusterlet/klusterlet-work-deployment.yaml
qiujian16: The env field will be change based on clusterName, I think that will trigger deployment. I will add a test to ensure that.

test/integration/klusterlet_test.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		_, err = kubeClient.CoreV1().Secrets(klusterletNamespace).Create(context.Background(), bootStrapSecret, metav1.CreateOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+		klusterlet = &operatorapiv1.Klusterlet{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("klusterlet-%s", rand.String(6)),
+			},
+			Spec: operatorapiv1.KlusterletSpec{
+				RegistrationImagePullSpec: "quay.io/open-cluster-management/registrationi",
```
pmorie: `registrationi`

manifests/klusterlet/klusterlet-work-deployment.yaml
pmorie: The deployment will be applied when the `spec.clusterName` field of the `Klusterlet` CR changes. So, it will switch correctly between:

1. `spec.clusterName=a`
2. `spec.clusterName=b`
3. `spec.clusterName` is unset
4. `spec.clusterName=c`

However, I am wondering whether the value will be correct when the system (3) - since the value inject will be the prior value set in the kubeconfig _before the pod starts_. So, I would expect when going from `spec.clusterName=a` to `spec.clusterName` is unset, that you would be injected with `a` when the pod started after the spec field for the name was unset.

Does that make sense?

qiujian16: yes, that makes sense. In this case, the registration agent will be set with unset cluster name, and generate a new cluster name, then saved to the secret. However, work agent is already started using the old cluster name. I think we have to reconcile again when the cluster name in secret is changed. So we cannot use env var to ref secret here.

pmorie:
e2e failure was an ci flake provisioning hub cluster

PR #33, opened 2020-06-15
Use generation to track update

qiujian16:


pkg/helpers/queukey.go
```diff
@@ -0,0 +1,119 @@
+package helpers
```
pmorie, line 1: typo in this file's name: queuekey.go

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -264,17 +270,51 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+		return
+	}
+
+	workDeploymentName := fmt.Sprintf("%s-work-agent", config.KlusterletName)
+	workDeployment, err := n.kubeClient.AppsV1().Deployments(config.KlusterletNamespace).Get(ctx, workDeploymentName, metav1.GetOptions{})
+	if err != nil {
+		return
+	}
+
+	if len(workDeployment.Spec.Template.Spec.Containers) != 1 {
```
pmorie, line 302: This might lead to some surprising behaviors when people manipulate the payload, do we need this check?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -264,17 +270,51 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+			oldStatus.ObservedGeneration = observedKlusterletGeneration
+			return nil
+		},
+	)
 	return nil
 }
 
+// Check if cluster name is changed in hub kubeconfig secret by comparing with deployment args. Force update the deployment
+// if the secret is changed.
+func (n *klusterletController) forceRollOutWorkAgent(ctx context.Context, config klusterletConfig, forceRollOut *bool) {
```
pmorie, line 291: My own personal preference here is to avoid mutating inputs and instead take an input parameter and output the result. Think it over, it can be done in a follow-up.

test/integration/clustermanager_test.go
```diff
@@ -169,52 +166,105 @@ var _ = ginkgo.Describe("ClusterManager", func() {
...
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
+
+		// Check if generations are correct
+		gomega.Eventually(func() bool {
+			actual, err := operatorClient.OperatorV1().ClusterManagers().Get(context.Background(), clusterManagerName, metav1.GetOptions{})
+			if err != nil {
+				return false
+			}
+
+			registrationoDeployment, err := kubeClient.AppsV1().Deployments(hubNamespace).Get(context.Background(), hubRegistrationDeployment, metav1.GetOptions{})
```
pmorie, line 257: typo: registrationoDeployment

pmorie (commented):
A couple nits and questions, but overall looks good

pmorie:
I think the scenarios covered here make sense to validate with an e2e. The integration tests are pretty thorough - I don't think we need e2e coverage of every single scenario covered by those tests, but a key scenario or two will be useful to validate against a real system. Maybe registration with and without cluster name change?

PR #35, opened 2020-06-15
Split Install of Operator and CR into 2 steps

zkayyali812:
This PR is to split the installation make targets into 2 optional targets.
- First to install the operator
- Second to apply the CR 

We aim to include the registration operator in our installer, and would prefer if we could separate this logic into the above steps to better fit our install processes.

--
This PR will allow us to install the operator, without installing the CR, so we can later drive the installation of the CR. This would only be for developer consumption.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,13 @@ ifeq ($(GOHOSTOS),darwin)
 	endif
 endif
 
+SED_CMD:=sed
+ifeq ($(GOHOSTOS),darwin)
+	ifeq ($(GOHOSTARCH),amd64)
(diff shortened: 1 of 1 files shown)


nathanweatherly (approved):

qiujian16:
@pmorie not quite sure why lgtm label is added, but i think we can use this pr to trigger postsubmit when build-harness is ready.

pmorie:
> not quite sure why lgtm label is added

Nathan approved the PR in the github UX, which is translated into lgtm tag if someone is in the org or a collaborator afaik.

PR #36, opened 2020-06-16
check managed cluster agent secrets status

skeeey:


pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
```diff
@@ -200,3 +297,143 @@ func (k *klusterletStatusController) sync(ctx context.Context, controllerContext
...
+			os.Remove(path.Join("/tmp", key))
+		}
+	}()
+
+	for key, data := range secret.Data {
+		if err := ioutil.WriteFile(path.Join("/tmp", key), data, 0644); err != nil {
+			return nil, err
+		}
+	}
+	restConfig, err := clientcmd.BuildConfigFromFlags("", path.Join("/tmp", "kubeconfig"))
```
pmorie: Instead of writing to the filesystem, use the [`Load`](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L406) method, which takes a byte array and gives you a config.

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
```diff
@@ -200,3 +297,143 @@ func (k *klusterletStatusController) sync(ctx context.Context, controllerContext
...
+	}
+	return kubernetes.NewForConfig(restConfig)
+}
+
+func createSelfSubjectAccessReviews(
+	ctx context.Context,
+	kubeClient kubernetes.Interface,
+	selfSubjectAccessReviews []authorizationv1.SelfSubjectAccessReview) (bool, *authorizationv1.SelfSubjectAccessReview, error) {
+	for _, subjectAccessReview := range selfSubjectAccessReviews {
+		ssar, err := kubeClient.AuthorizationV1().SelfSubjectAccessReviews().Create(ctx, &subjectAccessReview, metav1.CreateOptions{})
```
pmorie, line 326: The go `range` statement strikes again! It's not safe to use pointers to variables like `subjectAccessReview` that are created by the range statement. When you take a pointer to this, you are taking a pointer to _the variable you're using in the loop_, not the object in the collection. So, this will likely not work as you might expect.

Instead, you need to use this construction:

```go
for i := range selfSubjectAccessReviews {
  sar := selfSubjectAccessReviews[i]

  // use sar
  thing(&sar)
}
```

[more info](https://tam7t.com/golang-range-and-pointers/) here

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
```diff
@@ -200,3 +297,143 @@ func (k *klusterletStatusController) sync(ctx context.Context, controllerContext
...
+			return false, &subjectAccessReview, err
+		}
+		if !ssar.Status.Allowed {
+			return false, &subjectAccessReview, nil
+		}
+	}
+	return true, nil, nil
+}
+
+func getBootstrapSelfSubjectAccessReviews() []authorizationv1.SelfSubjectAccessReview {
```
pmorie, line 337: It would be awesome to be able to generate these somehow, but let's not bother for now.

pmorie (commented):
Mostly looking good, but let's avoid filesystem io and there's a bug you must correct w/ the range statement.

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
skeeey: I think we cannot use `Load` method here, for hub `kubeconfig`, its `client-certificate` and `client-key` are file path, rather than byte array, so if we use `Load` method, it will be failed, also for bootstrap `kubeconfig`, it is mounted from a secret, I think it also may be the file path, rather than the byte array

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
skeeey: fixed

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
pmorie: Okay - let's use ioutil.TempDir so that these tests don't collide with each other then

pmorie:
Just one remaining comment about `ioutil.TempDir`

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
skeeey: done

PR #37, opened 2020-06-16
Followup of pr33

qiujian16:
1. fix some typos
2. change forceRollOut func to be immutable


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -288,31 +290,25 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
-	if *forceRollOut {
-		return
-	}
-
+func (n *klusterletController) forceRollOutWorkAgent(ctx context.Context, config klusterletConfig) bool {
 	workDeploymentName := fmt.Sprintf("%s-work-agent", config.KlusterletName)
 	workDeployment, err := n.kubeClient.AppsV1().Deployments(config.KlusterletNamespace).Get(ctx, workDeploymentName, metav1.GetOptions{})
 	if err != nil {
-		return
+		return false
```
pmorie, line 298: shouldn't this be true? If we didn't find it, it should be rolled out, right?

pmorie (commented):
One question

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
qiujian16: I think we will create it in applyDeployment  if it is not found in the following. I will add a comment here.

PR #38, opened 2020-06-16
check hub deployment status

skeeey:


pkg/operators/clustermanager/controllers/statuscontroller/clustermanager_status_controller.go
```diff
@@ -0,0 +1,109 @@
+package clustermanagerstatuscontroller
```
pmorie: I would personally avoid package names that don't match the enclosing directory path, but you can use your own best judgement here.

test/integration/clustermanager_test.go
```diff
@@ -267,4 +267,28 @@ var _ = ginkgo.Describe("ClusterManager", func() {
 			return true
 		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
 	})
+
+	ginkgo.It("Cluster manager should have correct degraded conditions", func() {
```
pmorie: I don't understand how we can expect the deployment to exist, since there is no `BeforeSuite` creating a `ClusterManager`, but I feel I must be missing something about ginkgo here. I don't want to be pedantic and force you to use a `BeforeEach` if it's not required, will ping you later on slack about this so we can chat about it.

pkg/operators/clustermanager/controllers/statuscontroller/clustermanager_status_controller.go
skeeey: yes, this package name is a bit redundant, I use `statuscontroller` instead

test/integration/clustermanager_test.go
skeeey: OK, Indeed current test case is confused，the test cases
https://github.com/open-cluster-management/registration-operator/blob/master/test/integration/clustermanager_test.go#L172
https://github.com/open-cluster-management/registration-operator/blob/master/test/integration/clustermanager_test.go#L227 and 
new degrade test case are depend on https://github.com/open-cluster-management/registration-operator/blob/master/test/integration/clustermanager_test.go#L76 (it will create a `ClusterManager`)

the `ClusterManager` has only one fixed namespace in hub cluster (`open-cluster-management-hub`), and controller runtime envtest does not enable the garbage collector, so I think we cannot create `ClusterManager` on `BeforeEach` and delete the `ClusterManager` on `AfterEach`, this because when we delete `ClusterManager`, the cluster manager controller will delete the `ClusterManager` namespace, because there is no garbage collector, the namespace will keep the terminating status

I try to refactor the test cases
1. put the create `ClusterManager` action on `BeforeSuite`
2. use two context to organize the test cases

I think this will be more clear than before, I will ping you on slack

test/integration/clustermanager_test.go
```diff
@@ -167,104 +143,130 @@ var _ = ginkgo.Describe("ClusterManager", func() {
...
+				actualGeneration := helpers.FindGenerationStatus(actual.Status.Generations, deploymentGeneration)
+				if deploymentGeneration.LastGeneration != actualGeneration.LastGeneration {
+					return false
+				}
+				return true
+			}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
+		})
+	})
+
+	ginkgo.Context("klusterlet statuses", func() {
```
pmorie: shouldn't this be `Cluster manager statuses`?

test/integration/integration_suite_test.go
```diff
@@ -64,6 +68,18 @@ var _ = ginkgo.BeforeSuite(func(done ginkgo.Done) {
 	gomega.Expect(err).ToNot(gomega.HaveOccurred())
 	gomega.Expect(kubeClient).ToNot(gomega.BeNil())
 
+	// prepare a ClusterManager
```
pmorie, line 71: I think this is more appropriate for a BeforeEach in clustermanager_test.go

pmorie (commented):
Just a couple minor feedbacks

test/integration/integration_suite_test.go
pmorie: I'm sorry, I missed your comment earlier somehow, reading now.

test/integration/clustermanager_test.go
pmorie: > the `ClusterManager` has only one fixed namespace in hub cluster (`open-cluster-management-hub`), and controller runtime envtest does not enable the garbage collector, so I think we cannot create `ClusterManager` on `BeforeEach` and delete the `ClusterManager` on `AfterEach`, this because when we delete `ClusterManager`, the cluster manager controller will delete the `ClusterManager` namespace, because there is no garbage collector, the namespace will keep the terminating status

Ugh, you're right.

test/integration/clustermanager_test.go
skeeey: fixed

PR #39, opened 2020-06-17
add lease role for hub

zhiweiyin318:
I found failed e2e tests because the cluster-manager-registration-controller pod is forbidden to list lease.
So add lease role to allow hub to manage coordination.k8s.io/lease since registration has supported spoke cluster available check.


pmorie:
Looks like a potential flake of the webhook tests

/retest

PR #42, opened 2020-06-24
add e2e tests

zhiweiyin318:
use `make test-e2e` to start e2e tests. 

add 3 cases:
1. deploy a klusterlet CR with a specified managecluster name and approve the managedcluster.
2. deploy a klusterlet CR without managecluster name and approve the managedcluster.
3. approve a managecluster and deploy a configmap using manifestwork and then delete the klusterlet .


pmorie:
We will need to add the tests in this PR into the e2e job for this repo

test/e2e/common.go
```diff
@@ -0,0 +1,370 @@
...
+var (
+	kubeClient         kubernetes.Interface
+	clusterCfg         *rest.Config
+	operatorClient     operatorclient.Interface
+	clusterClient      clusterclient.Interface
+	workClient         workv1client.Interface
+	bootstrapHubSecret *corev1.Secret
+)
+
+func createKlusterlet(name, clusterName, agentNamespace string) (realClusterName string) {
```
pmorie: We should inject dependencies of this method instead of referring to package vars, especially if there is a thought to possibly export these methods in the future

test/e2e/common.go
```diff
@@ -0,0 +1,370 @@
...
+			}
+			_, err := kubeClient.CoreV1().Namespaces().Create(context.TODO(), &namespace, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		secret := bootstrapHubSecret.DeepCopy()
+		secret.ObjectMeta.ResourceVersion = ""
+		secret.SetNamespace(agentNamespace)
+		_, err = kubeClient.CoreV1().Secrets(agentNamespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
```
pmorie: It's better to just return errors from methods like these than to call `Expect`. You can call `Expect` in the calling method, and the method's code will be easier to read if it just returns errors when they occur.

test/e2e/common.go
```diff
@@ -0,0 +1,370 @@
...
+				{
+					URL: "https://localhost",
+				},
+			},
+			ClusterName: clusterName,
+			Namespace:   agentNamespace,
+		},
+	}
+
+	Expect(name).NotTo(BeEmpty())
```
pmorie: it's better to check input values before doing anything with them; this should be at the top of this method

test/e2e/common.go
```diff
@@ -0,0 +1,370 @@
...
+		secret.ObjectMeta.ResourceVersion = ""
+		secret.SetNamespace(agentNamespace)
+		_, err = kubeClient.CoreV1().Secrets(agentNamespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+	}
+
+	// create klusterlet CR
+	realKlusterlet, err := operatorClient.OperatorV1().Klusterlets().Create(context.TODO(), klusterlet, metav1.CreateOptions{})
+	Expect(err).NotTo(HaveOccurred())
+	realClusterName = realKlusterlet.Spec.ClusterName
```
pmorie: this is probably where your method should end, and the 'does the ManagedCluster exist?' concern factored into a separate method.

test/e2e/common.go
```diff
@@ -0,0 +1,370 @@
...
+
+	// approve csr
+	approveCSR(realClusterName)
+
+	// accept client
+	acceptsClient(realClusterName)
+	return
+}
+
+func approveCSR(clusterName string) {
```
pmorie: this should get injected w/ its dependencies (client etc)

test/e2e/common.go
```diff
@@ -0,0 +1,370 @@
...
+				},
+			},
+		},
+	}
+	_, err := workClient.WorkV1().ManifestWorks(clusterName).
+		Create(context.TODO(), manifestWork, metav1.CreateOptions{})
+	Expect(err).NotTo(HaveOccurred())
+}
+
+func cleanResources() {
```
pmorie: Be careful about this - we should be able to run the e2e against a live production system as part of a health check if we want to. As written, this will wipe a system :)

test/e2e/common.go
```diff
@@ -0,0 +1,370 @@
...
+
+func checkKlusterletOperatorReady() {
+	// make sure klusterlet operator deployment is created
+	Eventually(func() error {
+		_, err := kubeClient.AppsV1().Deployments(operatorNamespace).
+			Get(context.TODO(), klusterletOperator, metav1.GetOptions{})
+		return err
+	}, eventuallyTimeout, eventuallyInterval).Should(Succeed())
+
+	// make sure open-cluster-management-agent namespace is created
```
pmorie: why this specific namespace?

test/e2e/klusterlet_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/util/rand"
+)
+
+var _ = Describe("Test Cases: Create klusterlet", func() {
```
pmorie: You can remove the `test cases` and `sub case` verbiage from the descriptions here

pmorie (commented):
These look like very sensible test cases, thanks! A couple feedbacks on factoring of methods here that I would like you to address first.

test/e2e/common.go
zhiweiyin318: I refactored to a new object named `Tester` with dependencies and common methods.

test/e2e/common.go
zhiweiyin318: done

test/e2e/common.go
zhiweiyin318: done

test/e2e/common.go
zhiweiyin318: I split the old method into CreateKlusterlet, GetCreatedManagedCluster, ApproveCSR, AcceptsClient, CheckManagedClusterStatus.

test/e2e/common.go
zhiweiyin318: done with object `Tester`

test/e2e/common.go
zhiweiyin318: refactor to only delete resources created by e2e tests

test/e2e/common.go
zhiweiyin318: delete this specific namespace, and user can customize the klusterlet agent namespace

test/e2e/klusterlet_test.go
zhiweiyin318: done

deads2k:
it looks like a good starting point to me

/lgtm
/approve

PR #43, opened 2020-06-29
rename clusterrole and clusterrolebinding

zhiweiyin318:
related issue: https://github.com/open-cluster-management/backlog/issues/3025


manifests/cluster-manager/cluster-manager-registration-clusterrole.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: system:open-cluster-management:{{ .ClusterManagerName }}-registration-controller
```
qiujian16, line 4: i think  the format is something like open-cluster-management:registration:{{ .ClusterManagerName }}-controller ?

zhiweiyin318: done

PR #45, opened 2020-06-30
Handle kube 1.11

qiujian16:
https://github.com/open-cluster-management/backlog/issues/3168

Since in kube 1.11 (ocp 3.11), there is no escalate verb, so operator cannot create cluster role with wider permission than its own cluster role. This pr is to add clusterrolebinding to bind clusterrole registration and work to operator. So operator is able to deploy registration and work.

Also same as work agent who will not be able to create additional clusterrole from manifestwork.

### DIFF
--- a/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
+++ b/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/version"
 	appsinformer "k8s.io/client-go/informers/apps/v1"
 	coreinformer "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
@@ -29,6 +30,7 @@ import (
 	operatorapiv1 "github.com/open-cluster-management/api/operator/v1"
 	"github.com/open-cluster-management/registration-operator/pkg/helpers"
 	"github.com/open-cluster-management/registration-operator/pkg/operators/klusterlet/bindata"
+	"github.com/open-cluster-management/registration-operator/pkg/operators/klusterlet/kube111bindata"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


manifests/klusterletkube111/klusterlet-registration-operator-registration-clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+# Create the cluster rolebinding so klusterlet can deploy registration agent
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: open-cluster-management:registration-operator:{{ .KlusterletName }}-registration
```
deads2k: can you add a `-111` suffix or some such?

manifests/klusterletkube111/klusterlet-registration-operator-work-clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+# Create the cluster rolebinding so klusterlet can deploy work agent
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: open-cluster-management:registration-operator:{{ .KlusterletName }}-work
```
deads2k: same suffix comment here.

deads2k:
I'd like to see the names updated to be very clear when people are reading them.  But otherwise I can get behind this.  Thanks for the cleanup afterwards.

/approve

manifests/klusterletkube111/klusterlet-registration-operator-registration-clusterrolebinding.yaml
pmorie: +1

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -167,7 +183,25 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		return err
 	}
 
-	// Deploy the static resources
+	errs := []error{}
```
pmorie, line 185: I wonder how we can reliably test this later with e2e, hrm

pmorie: Maybe we can tag e2es as expected to work under 1.11 and put a CI workflow together that provisions 3.11 - but that might be some wishful thinking

pmorie (commented):
I'll be on later to help you get this in, I agree w/ the minor ask from David.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
qiujian16: i  think it is not easy to e2e this  right now...It will at least need two clusters, hub which is 4.x and spoke  which is 3.11...But i will do test manually.

qiujian16:
/hold
still need some test on ocp 3.11

manifests/klusterletkube111/klusterlet-registration-operator-registration-clusterrolebinding.yaml
qiujian16: fixed

manifests/klusterletkube111/klusterlet-registration-operator-work-clusterrolebinding.yaml
qiujian16: fixed

PR #50, opened 2020-07-10
add imagepullsecret

zhiweiyin318:
related issue: https://github.com/open-cluster-management/backlog/issues/3451


deploy/cluster-manager/crds/0000_01_operator.open-cluster-management.io_clustermanagers.crd.yaml
```diff
@@ -37,6 +37,10 @@ spec:
             that govern registration and work distribution for attached Klusterlets.
           type: object
           properties:
+            imagePullSecret:
+              description: ImagePullSecret represents the secret to pull images of
```
deads2k, line 41: I'm not clear on the value of allowing a user to customize the name of the secret containing the image pull credentials.  Why don't we simply declare that it will be secret `Foo`?  The secret exists in a namespace we own, so this seems like an area of flexibility you will have to investigate for every bug that doesn't add power that a user needs.

In contrast, some namespaces are user owned (like `openshift-config`), where transitions may have conflicting keys and the like and require flexibility.

manifests/cluster-manager/cluster-manager-registration-serviceaccount.yaml
```diff
@@ -3,3 +3,7 @@ kind: ServiceAccount
 metadata:
   name: {{ .ClusterManagerName }}-registration-controller-sa
   namespace: {{ .ClusterManagerNamespace }}
+{{if .ImagePullSecret}}
+imagePullSecrets:
+- name: {{ .ImagePullSecret }}
```
deads2k, line 8: I would expect missing imagepullsecret refs to be ignored, so if you avoid having a field, this manifest could just always list it.

deads2k:
spoke on slack. The operator can expect to see the imagepull secret in its namespace with a fixed name (perhaps open-cluster-management-pull-credentials). If that secret is present, it will be injected into every spot that needs it.

Since the operator itself may need it on its deployment to pull the operator image, the secret/open-cluster-management-pull-credentials can be unconditionally added to the operator serviceaccount's imagepullsecrets.  Since missing pull  secrets  are skipped, this is acceptable even if there is no pull secret.

This eliminate the need for a configuration API for exception cases and allows injection by end users and by controllers trying to install this.

PR #51, opened 2020-07-10
Sync pull secret

qiujian16:
related issue: open-cluster-management/backlog#3451

### DIFF
--- a/manifests/klusterlet/klusterlet-registration-serviceaccount.yaml
+++ b/manifests/klusterlet/klusterlet-registration-serviceaccount.yaml
@@ -3,3 +3,5 @@ kind: ServiceAccount
 metadata:
   name: {{ .KlusterletName }}-registration-sa
   namespace: {{ .KlusterletNamespace }}
+imagePullSecrets:
+- name: open-cluster-management-image-pull-credentials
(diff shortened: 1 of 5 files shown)


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -35,6 +35,7 @@ import (
 
 const (
 	klusterletFinalizer = "operator.open-cluster-management.io/klusterlet-cleanup"
+	imagePullSecret     = "pull-secret"
```
deads2k: I recommend a more descriptive name like `open-cluster-management-image-pull-credentials`

manifests/klusterlet/klusterlet-work-serviceaccount.yaml
```diff
@@ -3,3 +3,5 @@ kind: ServiceAccount
 metadata:
   name: {{ .KlusterletName }}-work-sa
   namespace: {{ .KlusterletNamespace }}
+imagePullSecrets:
+- name: {{ .ImagePullSecret }}
```
deads2k: this can be hardcoded, right?  No need to have a configuration option and make us look in two places.

manifests/klusterlet/klusterlet-registration-serviceaccount.yaml
```diff
@@ -3,3 +3,5 @@ kind: ServiceAccount
 metadata:
   name: {{ .KlusterletName }}-registration-sa
   namespace: {{ .KlusterletNamespace }}
+imagePullSecrets:
+- name: {{ .ImagePullSecret }}
```
deads2k: this can be hardcoded, right?  No need to have a configuration option and make us look in two places.

deads2k:
/approve

I'd like to see a different name and I recommend hardcoding the value.

PR #55, opened 2020-07-16
refactor degraded condition logic

skeeey:
refactor degraded condition logic to make the code more clear and fix below problems:

1. after the bootstrap, the registration agent keeps working, but the bootstrap secret will be expired, the registration degraded condition will prompt `BootstrapSecretUnauthorized`
2. if the hub cluster-admin approve the csr, but does not accept the managed cluster, at this time point, the registration and work agent degraded conditions are not consistent, registration agent degraded condition is `HubConfigSecretUnauthorized`, but work agent degraded condition  is `KubeConfigMissing`

### DIFF
--- a/pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
+++ b/pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"strings"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -41,6 +42,8 @@ type klusterletStatusController struct {
 
 const (
 	klusterletNamespace            = "open-cluster-management-agent"
+	klusterletRegistration         = "Registration"
+	klusterletWork                 = "Work"
 	klusterletRegistrationDegraded = "KlusterletRegistrationDegraded"
 	klusterletWorKDegraded         = "KlusterletWorkDegraded"
 )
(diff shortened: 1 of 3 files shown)


deads2k:
> 1. after the bootstrap, the registration agent keeps working, but the bootstrap secret will be expired, the registration degraded condition will prompt `BootstrapSecretUnauthorized`

If the bootstrap credential is expired and becomes unauthorized, then this cluster would be unable to rejoin if it's current credential expired.  This actually happens to kubelets in some cases with catastrophic results on cluster recovery when the kubelet is silently unable to rejoin the cluster.  It seems like we want to be degraded in that case so that the cluster can be put into a safe state.

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
```diff
@@ -167,40 +123,27 @@ func (k *klusterletStatusController) sync(ctx context.Context, controllerContext
...
-			registrationDegradedCondition.Status = metav1.ConditionTrue
-			registrationDegradedCondition.Reason = "ClusterNameMissing"
-			// Work condition will be the same as registration
-			workDegradedCondition.Message = registrationDegradedCondition.Message
-			workDegradedCondition.Status = registrationDegradedCondition.Status
-			workDegradedCondition.Reason = registrationDegradedCondition.Reason
-
+	// If hub kubeconfig does not exist, the bootstrap may not finish, check the bootstrap secret firstly
+	if hubConfigSecret.Data["kubeconfig"] == nil {
+		if registrationDegraded := k.checkBootstrapSecret(ctx, klusterletNS); registrationDegraded != nil {
```
deads2k: I think the bootstrap credentials should always be checked.  I agree that check should not prevent other checks.

deads2k:
What do you think about factoring this to list all the reason that the registration is degraded.  We do this in openshift by having multiple FooDegraded conditions, but you can avoid that big change for the moment by unioning a slice of Degraded conditions instead of exiting the loop at the first one.

deads2k:
> What do you think about factoring this to list all the reason that the registration is degraded. We do this in openshift by having multiple FooDegraded conditions, but you can avoid that big change for the moment by unioning a slice of Degraded conditions instead of exiting the loop at the first one.

For instance

```go

// returns nil if it is not degrade
type degradeCheckFunc(whatever you need) (*Condition)

degradedCheckFns := []degradeCheckFunc{
  bootstrapSecretExists
  bootstrapPermissionsCorrect
  hubKubeconfigExists
  clusterNameExists
}

degradedConditionReasons := []string{}
degradedConditionMessages := []string{}
for _, degradedCheckFn := range degradedCheckFns{
  currCond := degradedCheckFn(args)
  if currCond == nil{
     continue
  }
  //append
}

if len(degradedCondtiionReasons) == 0{
  // set degraded false
  return
}

degradedCondition.Status = true
degradedCondition.Reason = string.Join(degradedConditionReasons, ",")
degradedCondition.Message = strings.Join(degradedCondtiionMessages, "\n")


```

PR #59, opened 2020-07-20
output logs when e2e case fails

zhiweiyin318:
output logs when e2e case fails for debug.

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -1,8 +1,10 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"time"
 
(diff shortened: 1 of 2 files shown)


test/e2e/common.go
```diff
@@ -425,3 +427,69 @@ func (t *Tester) GetClusterNameFromKlusterlet(klusterletName string) (string, er
...
+	if err != nil {
+		klog.Errorf("failed to list klusterlets. error: %v", err)
+	}
+	for _, klusterlet := range klusterletes.Items {
+		klog.Infof("klusterlet %v : %#v \n", klusterlet.Name, klusterlet)
+	}
+
+	managedClusters, err := t.ClusterClient.ClusterV1().ManagedClusters().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("failed to list mangedClusters. error: %v", err)
```
pmorie: typo: managed

test/e2e/common.go
```diff
@@ -425,3 +427,69 @@ func (t *Tester) GetClusterNameFromKlusterlet(klusterletName string) (string, er
...
+		klog.Infof("klusterlet agent pod %v/%v : %#v \n", pod.Namespace, pod.Name, pod)
+		logs, err := t.PodLog(pod.Name, pod.Namespace, int64(10))
+		if err != nil {
+			klog.Errorf("failed to get pod %v/%v log. error: %v", pod.Namespace, pod.Name, err)
+			continue
+		}
+		klog.Infof("pod %v/%v logs:\n %v \n", pod.Namespace, pod.Name, logs)
+	}
+
+	manifestWorks, err := t.WorkClient.WorkV1().ManifestWorks("").List(context.TODO(), metav1.ListOptions{})
```
pmorie, line 479: This works for now, but you may want to track specific namespaces (ie, only dump from the namespaces involved in the tests) in situations where the e2e is running against a live system in a support situation in the future.

pmorie (commented):
Fix the typo and we'll land this. Thanks for the PR!

test/e2e/common.go
zhiweiyin318: done

test/e2e/common.go
zhiweiyin318: I add a TODO here, and will refactor this when I find a proper way to track the resources created during e2e.

PR #65, opened 2020-08-25
add clusterset crds

skeeey:


qiujian16:
I think you would want to add the two crds in hack/init.sh so we can verify the change on them.

skeeey:
added

PR #68, opened 2020-08-28
Delete ns of klusterlet

qiujian16:


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -399,7 +399,13 @@ func (n *klusterletController) cleanUp(ctx context.Context, controllerContext fa
 			}
 		}
 	}
-	return nil
+
+	// remove the klusterlet namespace
+	err = n.kubeClient.CoreV1().Namespaces().Delete(ctx, config.KlusterletNamespace, metav1.DeleteOptions{})
```
pmorie, line 404: note: verified that this should work correctly for default ns case

pmorie (commented):
I would probably suggest that we enhance the godoc for this type with information about the lifecycle of the ns; this PR lgtm

PR #77, opened 2020-09-15
using hub kubeconfig secret controller instead of mounting the secret

skeeey:


skeeey:
see https://github.com/open-cluster-management/backlog/issues/4992

PR #81, opened 2020-09-25
Remove apiservice for webhook

qiujian16:


qiujian16:
this is not needed to resolve webhook error any longer

PR #83, opened 2020-09-29
reconcile klusterlet bootstrap secret

skeeey:
see https://github.com/open-cluster-management/backlog/issues/4914

### DIFF
--- a/pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
+++ b/pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
@@ -1,0 +1,188 @@
+package bootstrapcontroller
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"strings"
+
+	operatorinformer "github.com/open-cluster-management/api/client/operator/informers/externalversions/operator/v1"
+	operatorlister "github.com/open-cluster-management/api/client/operator/listers/operator/v1"
+	"github.com/open-cluster-management/registration-operator/pkg/helpers"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	coreinformer "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
+	corelister "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/klog/v2"
+)
+
+// bootstrapController watches bootstrap hub kubeconfig secret, if the secret is changed with hub kube-apiserver ca or apiserver
+// endpoints, this controller will make the klusterlet re-bootstrap to get the new hub kubeconfig from hub cluster by deleting
+// the current hub kubeconfig secret and restart the klusterlet agents
+type bootstrapController struct {
+	kubeClient       kubernetes.Interface
+	klusterletLister operatorlister.KlusterletLister
+	secretLister     corelister.SecretLister
+}
+
+// NewBootstrapController returns a bootstrapController
+func NewBootstrapController(
+	kubeClient kubernetes.Interface,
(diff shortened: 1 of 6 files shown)


pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
```diff
@@ -0,0 +1,176 @@
...
+				return accessor.GetName()
+			}
+			return ""
+		}, secretInformer.Informer()).
+		ToController("BootstrapController", recorder)
+}
+
+func (k *bootstrapController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	bootstrapSecretName := controllerContext.QueueKey()
+	if bootstrapSecretName == "" {
```
qiujian16: should also know  the namespace of  bootstrap, and we can find the  related klusterlet by  namespace.

pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return nil
+}
+
+// reloadKlusterlet reload klusterlet by removing the current hub kubeconfig secret and klusterlet agent deployments
+func (k *bootstrapController) reloadKlusterlet(
+	ctx context.Context,
+	controllerContext factory.SyncContext,
+	klusterlet *operatorapiv1.Klusterlet) error {
+	namespace := klusterlet.Spec.Namespace
```
qiujian16: this can  be empty

pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
```diff
@@ -0,0 +1,176 @@
...
+	bootstrapHubKubeconfigSecret, err := k.secretLister.Secrets(namespace).Get(helpers.BootstrapHubKubeConfig)
```
qiujian16: i would get namespace from key and compare with klusterlet namespace, we do not need to get secret until we know there is a klusterlet deployed in this namespace.

qiujian16 (commented):
i think we need at least an integration test. This is a non-trivial change.

pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
skeeey: ok, I will get the namespace

pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
skeeey: get it

pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
skeeey: done

pkg/helpers/queuekey.go
```diff
@@ -105,6 +105,31 @@ func ClusterManagerDeploymentQueueKeyFunc(clusterManagerLister operatorlister.Cl
 	}
 }
 
+func BootstrapSecretQueueKeyFunc(klusterletLister operatorlister.KlusterletLister) factory.ObjectQueueKeyFunc {
```
qiujian16: i would put this in the controller since it seems only  bootstrap controller use it only.

pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
```diff
@@ -0,0 +1,158 @@
...
+		// the hub kubeconfig secret not found, could not have bootstrap yet, do nothing
+		return nil
+	case err != nil:
+		return err
+	}
+
+	hubKubeconfig, err := k.loadKubeConfig(hubKubeconfigSecret)
+	// the hub kubeconfig secret has errors, reload the klusterlet agents
+	if err != nil {
+		return k.reloadAgents(ctx, controllerContext, klusterletNamespace, klusterletName)
```
qiujian16: I‘d rather we do not reload here. An  empty hubkubeconfig secret can also trigger the reload, is it?

skeeey: do we have an empty hubkubecofnig? I want to add a judgement to judge the placeholder secret before here

pkg/helpers/queuekey.go
skeeey: done

pkg/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
skeeey: return here directly

PR #84, opened 2020-09-29
Disable cert rotation for temporary kube client

elgnay:
It's related to https://github.com/open-cluster-management/backlog/issues/5475.

### DIFF
--- a/pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
+++ b/pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
@@ -3,9 +3,6 @@ package statuscontroller
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
-	"os"
-	"path"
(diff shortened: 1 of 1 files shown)


pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
```diff
@@ -303,6 +303,19 @@ func buildKubeClientWithSecret(secret *corev1.Secret) (kubernetes.Interface, err
 	if err != nil {
 		return nil, err
 	}
+
+	// use embeded cert/key data instead of references to external cert/key files to
```
qiujian16: do we still need line  291-304? if we read from secret directly, i think you could  build restConfig using  byte  directly by  using clientcmd.Load()

elgnay: Code refactored as suggested.

PR #90, opened 2020-10-30
create a new Makefile target for spoke deploy on kind

mikeshng:
Signed-off-by: Mike Ng <ming@redhat.com>

create a new target that focuses on deploy spoke on kind.

As discussed with @qiujian16 this is needed for https://github.com/open-cluster-management/website/pull/4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,8 @@ OLM_VERSION?=0.16.1
 KUBECTL?=kubectl
 KUBECONFIG?=./.kubeconfig
 KLUSTERLET_KUBECONFIG_CONTEXT?=$(shell $(KUBECTL) config current-context)
+KLUSTERLET_KIND_KUBECONFIG?=$(HOME)/cluster1-kubeconfig
+HUB_KIND_KUBECONFIG?=$(HOME)/hub-kubeconfig
 KIND_CLUSTER?=kind
 
 OPERATOR_SDK_ARCHOS:=x86_64-linux-gnu
@@ -101,6 +103,10 @@ install-olm: ensure-operator-sdk
 	$(KUBECTL) get crds | grep clusterserviceversion ; if [ $$? -ne 0 ] ; then $(OPERATOR_SDK) olm install --version $(OLM_VERSION); fi
 	$(KUBECTL) get ns open-cluster-management ; if [ $$? -ne 0 ] ; then $(KUBECTL) create ns open-cluster-management ; fi
 
+install-olm-kind: ensure-operator-sdk
+	$(KUBECTL) --kubeconfig=$(KLUSTERLET_KIND_KUBECONFIG) get crds | grep clusterserviceversion ; if [ $$? -ne 0 ] ; then $(OPERATOR_SDK) olm install --version $(OLM_VERSION); fi
+	$(KUBECTL) --kubeconfig=$(KLUSTERLET_KIND_KUBECONFIG) get ns open-cluster-management ; if [ $$? -ne 0 ] ; then $(KUBECTL) create ns open-cluster-management ; fi
+
 deploy-hub: deploy-hub-operator apply-hub-cr
 
 deploy-hub-operator: install-olm munge-hub-csv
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -116,6 +122,9 @@ clean-hub: ensure-operator-sdk
 cluster-ip:
   CLUSTER_IP?=$(shell $(KUBECTL) get svc kubernetes -n default -o jsonpath="{.spec.clusterIP}")
 
+cluster-ip-kind:
```
qiujian16: cluster-hub-ip-kind?

Makefile
```diff
@@ -137,9 +153,17 @@ deploy-spoke: deploy-spoke-operator apply-spoke-cr
 deploy-spoke-operator: install-olm munge-spoke-csv bootstrap-secret
 	$(OPERATOR_SDK) run packagemanifests deploy/klusterlet/olm-catalog/klusterlet/ --namespace open-cluster-management --version $(CSV_VERSION) --install-mode SingleNamespace=open-cluster-management --timeout=10m
 
+deploy-spoke-kind: deploy-spoke-operator-kind apply-spoke-cr-kind
+
+deploy-spoke-operator-kind: install-olm-kind munge-spoke-csv bootstrap-secret-kind
+	$(OPERATOR_SDK) run packagemanifests deploy/klusterlet/olm-catalog/klusterlet/ --namespace open-cluster-management --version $(CSV_VERSION) --install-mode SingleNamespace=open-cluster-management --timeout=10m
+
 apply-spoke-cr:
 	$(SED_CMD) -e "s,quay.io/open-cluster-management/registration,$(REGISTRATION_IMAGE)," -e "s,quay.io/open-cluster-management/work,$(WORK_IMAGE)," deploy/klusterlet/config/samples/operator_open-cluster-management_klusterlets.cr.yaml | $(KUBECTL) apply -f -
```
qiujian16, line 162: we may not need  this sed command for now, since it is only used for e2e

Makefile
```diff
@@ -116,6 +122,9 @@ clean-hub: ensure-operator-sdk
 cluster-ip:
   CLUSTER_IP?=$(shell $(KUBECTL) get svc kubernetes -n default -o jsonpath="{.spec.clusterIP}")
 
+cluster-ip-kind:
+  CLUSTER_IP_KIND?=$(shell $(KUBECTL) --kubeconfig=$(HUB_KIND_KUBECONFIG) config view | grep server | awk '{ print $2 }' | cut -f3 -d/ | cut -f1 -d:)
```
qiujian16, line 126: it seems to me you do not need HUB_KIND_KUBECONFIG, KUBECONFIG is good enough?

Makefile
mikeshng: Thanks for the good catch. I will remove sed command.

Makefile
mikeshng: In my PR for the website, I actually changed exporting of the kubeconfig from hub to the managed one see: https://github.com/mikeshng/website/blob/update-foundation/content/getting-started/register-cluster.md#install-from-source

The main reason I did this is because of the olm install step doesn't seem to take in a kubeconfig and just uses whatever kubeconfig env is currently set. 
`$(OPERATOR_SDK) olm install --version $(OLM_VERSION)`
and during the deploy spoke we want to install the olm in the managed cluster.

So technically I actually didn't need to define `KLUSTERLET_KIND_KUBECONFIG` anymore. 

Reasons why I have both `HUB_KIND_KUBECONFIG` and `KLUSTERLET_KIND_KUBECONFIG` defined:
- the deploy spoke step is very complicated because involves interacting with the hub and the managed kubeconfig. having a clear variable makes the code a bit easier to understand.
- the `HUB_KIND_KUBECONFIG` is needed for the hub secret creation step
- I changed the instruction on the website to export KUBECONFIG as the managed one also makes it a bit easier to follow for the user since they are now prepping the managed cluster so it's more natural to export the kubeconfig as the managed one.

PR #96, opened 2020-11-30
Add crd & rbac rule for ClusterClaim controller

elgnay:

The task link: https://github.com/open-cluster-management/backlog/issues/7158


qiujian16:
do we remove the crd when klusterlet is deleted?

elgnay:
> do we remove the crd when klusterlet is deleted?

The logic is added to delete ClusterClaim crd once klusterlet is deleted.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -38,11 +38,13 @@ const (
 	klusterletFinalizer = "operator.open-cluster-management.io/klusterlet-cleanup"
 	imagePullSecret     = "open-cluster-management-image-pull-credentials"
 	klusterletApplied   = "Applied"
+	clusterClaimCrdFile = "manifests/klusterlet/0000_02_clusters.open-cluster-management.io_clusterclaims.crd.yaml"
```
qiujian16: should we just put this in staticResourceFiles?

elgnay: Yes, we can. I've updated the code as suggested.

PR #99, opened 2021-01-11
Add CertRotationController

elgnay:
The related issue: https://github.com/open-cluster-management/backlog/issues/8061


pkg/operators/clustermanager/controllers/certrotationcontroller/cabundle.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	caBundleConfigMap := originalCABundleConfigMap.DeepCopy()
+	if apierrors.IsNotFound(err) {
+		// create an empty one
+		caBundleConfigMap = &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: c.Namespace, Name: c.Name}}
+	}
+	if _, err = manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0]); err != nil {
+		return nil, err
+	}
+	if originalCABundleConfigMap == nil || originalCABundleConfigMap.Data == nil || !equality.Semantic.DeepEqual(originalCABundleConfigMap.Data, caBundleConfigMap.Data) {
```
qiujian16: it seems to me you do not need this if, it is ok to always apply

pkg/operators/clustermanager/controllers/certrotationcontroller/singer_test.go
```diff
@@ -0,0 +1,85 @@
+package certrotation
```
qiujian16: filename should be  signer_test.go

pkg/operators/clustermanager/controllers/certrotationcontroller/signer.go
```diff
@@ -0,0 +1,114 @@
+package certrotation
```
qiujian16: i would put signer/cabundle into a different dir from controller

pkg/operators/clustermanager/controllers/certrotationcontroller/certrotation_controller.go
```diff
@@ -0,0 +1,166 @@
...
+
+	klog.Infof("Reconciling ClusterManager %q", clustermanagers[0].Name)
+	// do nothing if the cluster manager is deleting
+	if !clustermanagers[0].DeletionTimestamp.IsZero() {
+		return nil
+	}
+
+	// create namespace for cluster manager if necessary
+	_, err = c.kubeClient.CoreV1().Namespaces().Get(ctx, helpers.ClusterManagerNamespace, metav1.GetOptions{})
+	switch {
```
qiujian16: how about we just  return err here? ns will created finally

pkg/operators/clustermanager/controllers/certrotationcontroller/target.go
```diff
@@ -0,0 +1,137 @@
...
+	if len(certificates) == 0 {
+		return "missing certificate"
+	}
+
+	cert := certificates[0]
+	if time.Now().After(cert.NotAfter) {
+		return "already expired"
+	}
+
+	//maxWait := cert.NotAfter.Sub(cert.NotBefore) / 5
```
qiujian16: remove  this?

pkg/operators/manager.go
```diff
@@ -44,6 +45,7 @@ func RunClusterManagerOperator(ctx context.Context, controllerContext *controlle
 	}
 
 	kubeInformer := informers.NewSharedInformerFactoryWithOptions(kubeClient, 5*time.Minute, informers.WithNamespace(helpers.ClusterManagerNamespace))
+	configmapInformer := kubeInformer.Core().V1().ConfigMaps()
```
qiujian16: directly put this in line 64

pkg/operators/clustermanager/controllers/certrotationcontroller/certrotation_controller.go
elgnay: I also updated clusterManagerController. It will not apply static resources, including the namespace, until the ca bundle configmap is created. So I have to create ns if it does not exist yet.

pkg/operators/clustermanager/controllers/certrotationcontroller/cabundle.go
elgnay: An event `CABundleUpdateRequired` will be created when an update is really needed. That's the reason why I have this if clause there.

pkg/operators/clustermanager/controllers/certrotationcontroller/singer_test.go
elgnay: Done.

pkg/operators/clustermanager/controllers/certrotationcontroller/signer.go
elgnay: Done. I also move the target into a new dir.

pkg/operators/clustermanager/controllers/certrotationcontroller/target.go
elgnay: Done.

pkg/operators/manager.go
elgnay: Done.

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -98,19 +94,22 @@ func NewClusterManagerController(
...
 		clusterManagerClient:  clusterManagerClient,
 		clusterManagerLister:  clusterManagerInformer.Lister(),
+		configMapLister:       configMapInformer.Lister(),
 		currentGeneration:     make([]int64, len(deploymentFiles)),
 	}
 
 	return factory.New().WithSync(controller.sync).
 		ResyncEvery(3*time.Minute).
 		WithInformersQueueKeyFunc(helpers.ClusterManagerDeploymentQueueKeyFunc(controller.clusterManagerLister), deploymentInformer.Informer()).
+		WithInformersQueueKeyFunc(helpers.ClusterManagerConfigmapQueueKeyFunc(controller.clusterManagerLister), configMapInformer.Informer()).
```
qiujian16: you would want to use a queukeyfunc with filter

pkg/operators/clustermanager/controllers/certrotationcontroller/certrotation_controller.go
qiujian16: i would apply  everything all  the time even  without a certain configmap. You finally get  into a steady  state, right? you may want to minimize coordination between two controllers.

pkg/operators/clustermanager/controllers/certrotationcontroller/cabundle.go
qiujian16: apply func will generate an event already.

pkg/operators/clustermanager/controllers/certrotationcontroller/certrotation_controller.go
elgnay: > i would apply everything all the time even without a certain configmap. You finally get into a steady state, right? you may want to minimize coordination between two controllers.

It makes sense.  Updated as suggested.

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
elgnay: Done.

pkg/operators/clustermanager/controllers/certrotationcontroller/cabundle.go
elgnay: apply func generates event once resource is updated. The event I mentioned above is created before change is made. The similar events are created for signing rotation and target rotation as well.

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -175,24 +182,20 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
...
 	}
-	config.RegistrationAPIServiceCABundle = base64.StdEncoding.EncodeToString(ca)
-	config.RegistrationServingCert = base64.StdEncoding.EncodeToString(cert)
-	config.RegistrationServingKey = base64.StdEncoding.EncodeToString(key)
-
-	ca, cert, key, err = n.ensureServingCertAndCA(
-		ctx, helpers.ClusterManagerNamespace, workWebhookCert, workWebhookService)
-	if err != nil {
-		return err
+	if err == nil {
```
qiujian16: how about
switch {
case erros.IsNotFound(err):
// do nothing
case err != nil:
  return  err
default:
  //set cabundle
}

elgnay: Updated.

PR #103, opened 2021-03-22
Enable support for "getting started" with kind on OS X

### LINKED ISSUE #102: Enable "getting-started" scenario with kind on OS X
Currently, the [getting started instructions](https://open-cluster-management.io/getting-started) for creating a hub cluster and managed cluster with kind do not work for OS X.  

This is due to the fact that the docker networking works differently from linux on OS X. The main issue is that it is not possible to reach directly a docker address on the docker networks such as `172.18.0.2` from the host on OS X. On linux those addresses are reachable, but there are still issues if using a version of kind > v 0.7 as pointed out in [this issue](https://github.com/kubernetes-sigs/kind/issues/1916) because kind > v0.7 uses host names instead than IP addresses and hosts name resolution does not work on the linux host for internal docker addresses.

pdettori:
Fixes #102

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,8 @@ KLUSTERLET_KUBECONFIG_CONTEXT?=$(shell $(KUBECTL) config current-context)
 KLUSTERLET_KIND_KUBECONFIG?=$(HOME)/cluster1-kubeconfig
 HUB_KIND_KUBECONFIG?=$(HOME)/hub-kubeconfig
 KIND_CLUSTER?=kind
+MANAGED_CLUSTER?=cluster1
+HUB_CLUSTER?=hub
 
 OPERATOR_SDK_ARCHOS:=x86_64-linux-gnu
 ifeq ($(GOHOSTOS),darwin)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -119,22 +122,23 @@ cluster-ip:
   CLUSTER_IP?=$(shell $(KUBECTL) get svc kubernetes -n default -o jsonpath="{.spec.clusterIP}")
 
 cluster-hub-ip-kind:
-  CLUSTER_IP_KIND?=$(shell $(KUBECTL) --kubeconfig=$(HUB_KIND_KUBECONFIG) config view | grep server | awk '{ print $2 }' | cut -f3 -d/ | cut -f1 -d:)
+  	CLUSTER_IP_KIND?=$(shell $(KUBECTL) --kubeconfig $(HOME)/hub-kubeconfig config view | grep server | awk '{ print $2 }' | cut -f3 -d/)
 
 bootstrap-secret: cluster-ip
 	cp $(KUBECONFIG) dev-kubeconfig
-	$(KUBECTL) config use-context $(KLUSTERLET_KUBECONFIG_CONTEXT)
+	$(KUBECTL) config use-context kind-$(MANAGED_CLUSTER)
```
qiujian16: I think this will affect our e2e test.

Makefile
```diff
@@ -40,6 +40,8 @@ KLUSTERLET_KUBECONFIG_CONTEXT?=$(shell $(KUBECTL) config current-context)
 KLUSTERLET_KIND_KUBECONFIG?=$(HOME)/cluster1-kubeconfig
 HUB_KIND_KUBECONFIG?=$(HOME)/hub-kubeconfig
 KIND_CLUSTER?=kind
+MANAGED_CLUSTER?=cluster1
```
qiujian16, line 42: What is the diff between MANAGED_CLUSTER and KIND_CLUSTER?

Makefile
pdettori: right, I see that now. How can I run the e2e test locally to debug ? `make test-e2e` ?

Makefile
pdettori: They are actually the same. But I think it would be clearer to just use `MANAGED_CLUSTER` as `KIND_CLUSTER` is ambiguous (the hub cluster is also a kind cluster). If you are ok with that I will just use `MANAGED_CLUSTER` and get rid of `KIND_CLUSTER`.

Makefile
qiujian16: yeah, you can run make test-e2e locally

Makefile
qiujian16: i think we can  use MANAGED_CLUSTER only

pdettori:
@qiujian16 I have updated the PR and now e2e tests are passing. I have also removed `KIND_CLUSTER`

qiujian16:
thanks, it looks good. Would you also update the README by removing the KIND_CLUSTER?

/approve

pdettori:
> thanks, it looks good. Would you also update the README by removing the KIND_CLUSTER?

Done. I also added the step to create the kind cluster and get the kubeconfig file in ./.kubeconfig as the `make deploy` was failing without that file being in place. It works fine now.

qiujian16:
thanks!

/lgtm

PR #116, opened 2021-05-05
Add management workload annotations

imiller0:
In support of the workload partitioning feature
(openshift/enhancements#703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

Signed-off-by: Ian Miller <imiller@redhat.com>

### DIFF
--- a/deploy/klusterlet/config/operator/namespace.yaml
+++ b/deploy/klusterlet/config/operator/namespace.yaml
@@ -1,4 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
+  annotations:
+    workload.openshift.io/allowed: "management"
   name: open-cluster-management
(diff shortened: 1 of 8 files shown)


imiller0:
@qiujian16 Can you please review this PR. The annotations need to be added to namespaces and pods created on the spoke cluster. With respect to that, one question in specific is whether the change in pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go is actually necessary?
Thanks!

qiujian16:
@imiller0 i think it is needed,  thanks!

qiujian16:
@hanqiuzh FYI  i think we need to add this annotation in rcm-controller also?

hanqiuzh:
Yes managedcluster-addon-controller (former rcm-controller) got one already: https://github.com/open-cluster-management/managedcluster-import-controller/pull/46

PR #117, opened 2021-05-07
Adjust agent replica based on infra api

qiujian16:
https://github.com/open-cluster-management/backlog/issues/10647


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -525,6 +534,36 @@ func (n *klusterletController) cleanUpAppliedManifestWorks(ctx context.Context,
 	return operatorhelpers.NewMultiLineAggregate(errs)
 }
 
+// determineReplica determines the replica of deployment based on:
+// 1. Read ControlPlaneTopology from infrastructure API. Return 1 if it is SingleReplicaTopologyMode, otherwise return 3.
+// 2. If infrastructure API does not exist, list master nodes in the cluster and return 1 if
+// the number of master nodes is equal or less than 1. Return 3 otherwise.
+func (n *klusterletController) determineReplica(ctx context.Context) int32 {
```
deads2k, line 541: This should be a check you perform once every ten minutes.  Depending on what kube you're running on top of, the number of nodes can change.  Ten minutes seems reactive enough to handle this, while being slow enough to avoid excessive listing.

deads2k: > This should be a check you perform once every ten minutes. Depending on what kube you're running on top of, the number of nodes can change. Ten minutes seems reactive enough to handle this, while being slow enough to avoid excessive listing.

Nevermind, I see you're doing exactly that.

PR #121, opened 2021-05-18
change makefile to support customizing hub name when deploying hub

xuezhaojun:


xuezhaojun:
The issues solved by another pr

PR #133, opened 2021-06-23
Fix no managecluster after deployment and modify readme

xuezhaojun:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,10 +46,7 @@ kustomize_dir:=$(dir $(KUSTOMIZE))
 
 KUBECTL?=kubectl
 KUBECONFIG?=./.kubeconfig
-KLUSTERLET_KUBECONFIG?=$(HOME)/cluster1-kubeconfig
-HUB_KUBECONFIG_CONTEXT?=$(shell $(KUBECTL) config current-context)
-HUB_KUBECONFIG?=$(HOME)/hub-kubeconfig
-MANAGED_CLUSTER?=cluster1
+HUB_KUBECONFIG=./.hub-kubeconfig
 
 OPERATOR_SDK_ARCHOS:=x86_64-linux-gnu
 ifeq ($(GOHOSTOS),darwin)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -106,9 +102,7 @@ clean-hub: ensure-operator-sdk
 	$(OPERATOR_SDK) cleanup cluster-manager --namespace open-cluster-management --timeout 10m
 
 cluster-ip:
-	cp $(KUBECONFIG) $(HUB_KUBECONFIG)
-	$(KUBECTL) config use-context $(HUB_KUBECONFIG_CONTEXT) --kubeconfig $(HUB_KUBECONFIG)
-	$(KUBECTL) config set clusters.$(HUB_KUBECONFIG_CONTEXT).server https://$(shell $(KUBECTL) get svc kubernetes -n default -o jsonpath="{.spec.clusterIP}") --kubeconfig $(HUB_KUBECONFIG)
+	$(KUBECTL) config set clusters.$(shell $(KUBECTL) config current-context --kubeconfig $(HUB_KUBECONFIG)).server https://$(shell $(KUBECTL) get svc kubernetes -n default -o jsonpath="{.spec.clusterIP}" --kubeconfig $(HUB_KUBECONFIG)) --kubeconfig $(HUB_KUBECONFIG)
```
qiujian16: dup --kubeconfig flag

Makefile
```diff
@@ -106,9 +102,7 @@ clean-hub: ensure-operator-sdk
 	$(OPERATOR_SDK) cleanup cluster-manager --namespace open-cluster-management --timeout 10m
 
 cluster-ip:
-	cp $(KUBECONFIG) $(HUB_KUBECONFIG)
```
qiujian16, line 109: I think we need to keep this otherwise we cannot run e2e automation in prow.

xuezhaojun: moved 'cp' into target `hub-kubeconfig`
testing it with e2e currently.

Makefile
xuezhaojun: Run without `--kubeconfig` cause no error, but with this flag means all things happen without any effects from other files.

For example:
I run `make deploy-hub` and set hub-kubeconfig, somehow I switched to another context.
Then I run `make cluster-ip` and get error, because `current-context` is not `hub-context`.
But with `--kubeconfig`, this error would never happened.

Makefile
xuezhaojun: e2e test is done.

xuezhaojun (commented):
Done

qiujian16 (commented):
/approve
/lgtm

PR #149, opened 2021-09-22
create migration requests for clusterset api

elgnay:
The related task: https://github.com/open-cluster-management/backlog/issues/16156


qiujian16 (commented):
could we move these to a different controllers (CRD migration controller)? We only need this in one release, and I would want the code to be separated.

elgnay:
> could we move these to a different controllers (CRD migration controller)? We only need this in one release, and I would want the code to be separated.

A separated controller is created for CRD migration.

PR #152, opened 2021-09-24
fix resource pressure on hub when too many managed cluster registrate at once

xuezhaojun:

### DIFF
--- a/pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
+++ b/pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
@@ -1,0 +1,435 @@
+package ssarcontroller
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	authorizationv1 "k8s.io/api/authorization/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	coreinformer "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
+	corelister "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	operatorv1client "open-cluster-management.io/api/client/operator/clientset/versioned/typed/operator/v1"
+	operatorinformer "open-cluster-management.io/api/client/operator/informers/externalversions/operator/v1"
+	operatorlister "open-cluster-management.io/api/client/operator/listers/operator/v1"
+	"open-cluster-management.io/registration-operator/pkg/helpers"
+)
+
+type ssarController struct {
+	kubeClient       kubernetes.Interface
+	secretLister     corelister.SecretLister
+	klusterletClient operatorv1client.KlusterletInterface
+	klusterletLister operatorlister.KlusterletLister
+	*klusterletLocker
+}
+
+type klusterletLocker struct {
+	sync.RWMutex
+	klusterletInChecking map[string]struct{}
+}
+
+const (
+	klusterletNamespace     = "open-cluster-management-agent"
+	bootstrapSecret         = "BootstrapSecret"
+	bootstrapSecretDegraded = "BootstrapSecretDegraded"
+	hubConfigSecret         = "HubConfigSecret"
+	hubConfigSecretDegraded = "HubConfigSecretDegraded"
+)
+
+func NewKlustrletSSARController(
+	kubeClient kubernetes.Interface,
+	klusterletClient operatorv1client.KlusterletInterface,
+	klusterletInformer operatorinformer.KlusterletInformer,
+	secretInformer coreinformer.SecretInformer,
+	recorder events.Recorder,
+) factory.Controller {
+	controller := &ssarController{
+		kubeClient:       kubeClient,
+		klusterletClient: klusterletClient,
+		klusterletLister: klusterletInformer.Lister(),
+		secretLister:     secretInformer.Lister(),
+		klusterletLocker: &klusterletLocker{
+			klusterletInChecking: make(map[string]struct{}),
+		},
+	}
+
+	return factory.New().WithSync(controller.sync).
+		WithInformersQueueKeyFunc(helpers.KlusterletSecretQueueKeyFunc(controller.klusterletLister), secretInformer.Informer()).
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
+			accessor, _ := meta.Accessor(obj)
+			return accessor.GetName()
+		}, klusterletInformer.Informer()).
+		ToController("KlusterletSSARController", recorder)
+}
+
+func (l *klusterletLocker) inSSARChecking(klusterletName string) bool {
+	l.RLock()
+	defer l.RUnlock()
+	_, ok := l.klusterletInChecking[klusterletName]
+	return ok
+}
+
+func (l *klusterletLocker) addSSARChecking(klusterletName string) {
+	l.Lock()
+	defer l.Unlock()
+	l.klusterletInChecking[klusterletName] = struct{}{}
+}
+
+func (l *klusterletLocker) deleteSSARChecking(klusterletName string) {
+	l.Lock()
+	defer l.Unlock()
+	delete(l.klusterletInChecking, klusterletName)
+}
+
+func (c *ssarController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	klusterletName := controllerContext.QueueKey()
+	if klusterletName == "" {
+		return nil
+	}
+	klusterlet, err := c.klusterletLister.Get(klusterletName)
+	switch {
+	case errors.IsNotFound(err):
+		return nil
+	case err != nil:
+		return err
+	}
+	klusterlet = klusterlet.DeepCopy()
+
+	// if the ssar checking is already processing, ignore reconciling this turn.
+	if c.inSSARChecking(klusterletName) {
+		klog.V(4).Info("Reconciling Klusterlet %q is already processing now", klusterletName)
+		return nil
+	}
+
+	c.addSSARChecking(klusterletName)
+	go func() {
+		defer c.deleteSSARChecking(klusterletName)
+
+		klog.V(4).Infof("Reconciling Klusterlet %q", klusterletName)
+		klusterletNS := klusterlet.Spec.Namespace
+		if klusterletNS == "" {
+			klusterletNS = klusterletNamespace
+		}
+
+		bootstrapDegradedCondition := checkAgentDegradedCondition(
+			ctx, c.kubeClient,
+			bootstrapSecret, bootstrapSecretDegraded,
+			klusterletAgent{
+				clusterName: klusterlet.Spec.ClusterName,
+				namespace:   klusterletNS,
+			},
+			[]degradedCheckFunc{checkBootstrapSecret},
+		)
+
+		hubConfigDegradedCondition := checkAgentDegradedCondition(
+			ctx, c.kubeClient,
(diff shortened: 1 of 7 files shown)


xuezhaojun:
refers to : https://github.com/open-cluster-management/backlog/issues/14614

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
```diff
@@ -296,8 +298,12 @@ func createSelfSubjectAccessReviews(
 	ctx context.Context,
 	kubeClient kubernetes.Interface,
 	selfSubjectAccessReviews []authorizationv1.SelfSubjectAccessReview) (bool, *authorizationv1.SelfSubjectAccessReview, error) {
+	// delay a random seconds(between 1-10) for every time creation, because too many managed clusters registration on hub and send ssar requests at once could cause resource pressure
+	time.Sleep(time.Second * time.Duration(rand.Intn(10)))
```
skeeey: how about delay the each request? on the request loop

```
for i := range selfSubjectAccessReviews {
 
}
```

skeeey: or we may use a delay queue to handle this

https://github.com/kubernetes/client-go/blob/master/util/workqueue/delaying_queue.go

we add the each ssar request to a delay queue and handle them

zhiweiyin318:
how much QPS has been reduced by using this solution vs. not using?

skeeey:
> how much QPS has been reduced by using this solution vs. not using?

This cannot reduce the QPS, it just make the requests become gently

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
xuezhaojun: Assuming we are using delay queue like the following example, the concern is we can not get the items in waitToAddQueue, we can only get items length of waitToProcessQueue.

```go
for i := range ssars {
  // ...
  dq.AddAfter(ssar[i], getRandomTime()) // dq is a delayqueue
}

for dq.Len() != 0 { // The Len() can only return the length of waitToProcessQueue, but we actually need the length of waitToAddQueue
   ssar := dq.Get()

  createRequest(ssar)

   dq.Done(item)
}
```

xuezhaojun: Delay each request also works, but I'm not sure what a proper random time range should I set.
Right now, I set it to be a random number between 1-1000, and the random delay time should be 1ms - 1000ms.

xuezhaojun:
https://coreos.slack.com/archives/D02G817B7DE/p1632487166003800
We could have a 1k K8S env for testing sooner.

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
skeeey: I think we need get the proper random time range with our test

if using delay queue, I think you need two steps

1. add the ssar to the queue
2. start a go routine to handle the queue

qiujian16: I would think you have a separated controller to handle sar only instead of in the status controller. It might be also reasonable to put it in the different condition from registration degraded.

xuezhaojun: @qiujian16 create a new controller `klusterlet_hubconfig_controller` without delete `status_controller` yet. Please have a review of this new controller, if it is written as expected, then I can add testcases and remove duplicate part in status controller.

@skeeey Tried to start a go routine, but it still has the issue mentioned above --- Because items are added after a while, so we can't know for sure : at a certain moment, whether all ssar are done or not.
Unless, we use `waitgroup` or `update condition for each ssar indenpendenly`, but it will also import complexity.

pkg/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
```diff
@@ -296,8 +298,13 @@ func createSelfSubjectAccessReviews(
 	ctx context.Context,
 	kubeClient kubernetes.Interface,
 	selfSubjectAccessReviews []authorizationv1.SelfSubjectAccessReview) (bool, *authorizationv1.SelfSubjectAccessReview, error) {
+
 	for i := range selfSubjectAccessReviews {
+		// delay a random seconds(between 1-1000ms) for every time creation, because too many managed clusters registration on hub and send ssar requests at once could cause resource pressure
```
qiujian16: I think you need to remove the original code in klusterlet_status_controller.go?

pkg/operators/klusterlet/controllers/hubconfigcontroller/klusterlet_hubconfig_controller.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+func createSelfSubjectAccessReviews(
+	ctx context.Context,
+	kubeClient kubernetes.Interface,
+	selfSubjectAccessReviews []authorizationv1.SelfSubjectAccessReview) (bool, *authorizationv1.SelfSubjectAccessReview, error) {
+
+	for i := range selfSubjectAccessReviews {
+		// delay a random seconds(between 1-1000ms) for every time creation, because too many managed clusters registration on hub and send ssar requests at once could cause resource pressure
+		time.Sleep(time.Millisecond * time.Duration(rand.Intn(1000)))
```
qiujian16: one of the main reason to have a separate controller is to avoid sleep here. Would you use jitter for the controller, instead of a sleep here?

xuezhaojun:
Update
1. using independent goroutine to process ssar, ignore reconcile when process are on going
2. reduce qps and burst to lowest to separate requests into a wider time range.

Doning:
1. delete duplicate part in status controller
2. add testcases

xuezhaojun:
Update：
1. status_controller.go:
    1. delete secret-related apart, only reconcile deployments
2. status_controller_test.go:
    1. only keeps test cases of deployments
3. ssar_controler.go:
    1. Condition Type:  change from `Registration`, `Work` to `BootstrapSecret`, `HubConfigSecret`
    2. Also Merge function `getRegistrationSelfSubjectAccessReviews` (to get ssar of `Work`) and `getRegistrationSelfSubjectAccessReviews` (to get ssar of `Registration`) to one fuction: `getHubConfigSSARs`
    3. Put ssar creation in a separated goroutine
    4. Limit QPS and Burst when `buildKubeClientWithSecret`
4. ssar_controler_test.go:
    1. Add corresponding test cases
5. klusterlet_test.go:
    1. modify to apply changes in controllers

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+func buildKubeClientWithSecret(secret *corev1.Secret) (kubernetes.Interface, error) {
+	restConfig, err := helpers.LoadClientConfigFromSecret(secret)
+	if err != nil {
+		return nil, err
+	}
+
+	// reduce qps and burst of client, because too many managed clusters registration on hub and send ssar requests at once could cause resource pressure
+	restConfig.QPS = 1
```
qiujian16: I  think we can make it to 5

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
```diff
@@ -0,0 +1,429 @@
...
+		WithInformersQueueKeyFunc(helpers.KlusterletSecretQueueKeyFunc(controller.klusterletLister), secretInformer.Informer()).
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
+			accessor, _ := meta.Accessor(obj)
+			return accessor.GetName()
+		}, klusterletInformer.Informer()).
+		ToController("KlusterletSSARController", recorder)
+}
+
+func (c *ssarController) inSSARChecking(klusterletName string) bool {
+	c.Lock()
```
qiujian16: this should be RLock?

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
```diff
@@ -0,0 +1,429 @@
...
+	operatorlister "open-cluster-management.io/api/client/operator/listers/operator/v1"
+	"open-cluster-management.io/registration-operator/pkg/helpers"
+)
+
+type ssarController struct {
+	kubeClient       kubernetes.Interface
+	secretLister     corelister.SecretLister
+	klusterletClient operatorv1client.KlusterletInterface
+	klusterletLister operatorlister.KlusterletLister
+	sync.RWMutex
```
qiujian16, line 38: i think you would want a lock for each klusterlet

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
xuezhaojun: Sure, but the default QPS is 5, should we set it to 5 explicitly?

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
xuezhaojun: Sure, it should be RLock and RUnLock

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
xuezhaojun: A little confused about this, it doesn't seem different klusterlet would have an effect on each other.

qiujian16: when you have another klusterlet comes in, you will get hung when add, right?

xuezhaojun: It won't hang because the lock only happens in the internal of `inSSARChecking`, `addSSARChecking`, `deleteSSARChecking`. The three functions are very basic so there would be no hang inside.

When cluster1 is processing, the  `addSSARChecking`, `deleteSSARChecking` are not calling obviously. 

The `inSSARChecking` is called when a new reconcile begins, but **immediately unlocked** and returns a result (for same klusterlet return true and for another no-processing klusterlet return false) and there would be no hanging stage.

I don't know if there is something I'm missing, I'm working on adding a test for this case.

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
xuezhaojun: And also may cause the jitter works too lite and not as we expected.
Still, this can be modified based on test results later. Currently set to 5.

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
xuezhaojun: Separate the locker into an independent struct and add a test function `TestLocker` to simulate the situation.

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller_test.go
```diff
@@ -0,0 +1,331 @@
...
+
+			testinghelper.AssertEqualNumber(t, len(operatorActions), 2)
+			testinghelper.AssertGet(t, operatorActions[0], "operator.open-cluster-management.io", "v1", "klusterlets")
+			testinghelper.AssertAction(t, operatorActions[1], "update")
+			testinghelper.AssertOnlyConditions(t, operatorActions[1].(clienttesting.UpdateActionImpl).Object, c.expectedConditions...)
+		})
+	}
+}
+
+func TestLocker(t *testing.T) {
```
xuezhaojun, line 273: @qiujian16  Plz have a view. This would depict klusterlets would not affect each other.

qiujian16 (commented):
/approve

xuezhaojun:
@skeeey  could you help with a lgtm tag?

PR #160, opened 2021-11-08
reduce the sar request when hub kubeconfig is valid

skeeey:
Signed-off-by: liuwei <liuweixa@redhat.com>

### DIFF
--- a/pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
+++ b/pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -26,6 +27,9 @@ import (
 	"open-cluster-management.io/registration-operator/pkg/helpers"
 )
 
+// SSARReSyncTime is exposed so that integration tests can crank up the controller sync speed.
+var SSARReSyncTime = 30 * time.Second
+
 type ssarController struct {
 	kubeClient       kubernetes.Interface
 	secretLister     corelister.SecretLister
@@ -45,6 +49,7 @@ const (
 	bootstrapSecretDegraded = "BootstrapSecretDegraded"
 	hubConfigSecret         = "HubConfigSecret"
 	hubConfigSecretDegraded = "HubConfigSecretDegraded"
+	hubConnectionDegraded   = "HubConnectionDegraded"
 )
 
 func NewKlustrletSSARController(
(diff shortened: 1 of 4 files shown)


skeeey:
refer to https://github.com/open-cluster-management-io/community/issues/99

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
```diff
@@ -106,14 +110,17 @@ func (c *ssarController) sync(ctx context.Context, controllerContext factory.Syn
 	}
 	klusterlet = klusterlet.DeepCopy()
 
-	// if the ssar checking is already processing, ignore reconciling this turn.
+	// if the ssar checking is already processing, requeue it.
```
qiujian16: why we need to requeue it? I think this is to avoid multiple checks happens

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
```diff
@@ -122,36 +129,60 @@ func (c *ssarController) sync(ctx context.Context, controllerContext factory.Syn
...
+				Status:  metav1.ConditionTrue,
+				Reason:  bootstrapDegradedCondition.Reason + "," + hubConfigDegradedCondition.Reason,
+				Message: bootstrapDegradedCondition.Message + "\n" + hubConfigDegradedCondition.Message,
+			}),
+		)
+		return err
+	})
 
-	return nil
+	return errs.Wait()
```
qiujian16: this makes it locked until error returns? I think we want to ensure that when a klusterlet comes, it spawns a goroutine and not wait anymore. So the check in this gorountine can be pretty slow due to the set on qps.

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
skeeey: this is for config secret update case, e.g. in our integration test,  we create the hub config secret then update it, for sar controller, when it handling the create event, the update event is coming, in this time, the update event will not be handled, so the update event will be missed

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
skeeey: yes, this will wait the error return, but I think this will not affect the creation of  goroutine, because when a klusterlet comes, there will be an event, this event will trigger a new call to the sync method, am I right?

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
skeeey: increase the requeue time

pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
skeeey: changed to when error is happened, we requeue the request

Issue #158, opened 2021-10-28
Allow klusterlet to run outside of managed cluster


qiujian16:
We should allow klusterlet (and all addons) to run outside of managed cluster.

This is useful when we want to:
1. run klusterlet on the hub cluster.
2. run klusterlet on a different cluster due to firewall setup.

/assign @zhujian7

zhujian7:
As of now, klusterlet can be run outside the managed cluster, but addons still can not.
I think the addon part should not be in release v0.6

zhujian7:
Proposal: https://github.com/open-cluster-management-io/enhancements/pull/36

qiujian16:
I think this has been finished

/close

PR #172, opened 2021-11-17
support deploy klusterlet outside of managed cluster

zhujian7:
Related PRs:
- https://github.com/open-cluster-management-io/api/pull/114
- https://github.com/open-cluster-management-io/registration/pull/175
- https://github.com/open-cluster-management-io/work/pull/104

related issue: open-cluster-management-io/registration-operator#158


zhujian7:
@qiujian16  @zhiweiyin318 @xuezhaojun @ianzhang366 This PR Is ready for review.

but still need to:
~~- [ ] change the namespace from `klusterlet name` to `{klusterlet name}-open-cluster-management-agent`~~
- [x] add UT, integration tests

manifests/klusterlet/detached/klusterlet-registration-rolebinding-management.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
```
qiujian16: why do we need to separated these. I think the manifests has only nuance with non-detached mode that you can render. Maintaining two lists of manifests would be hard since it is easy to update one and miss another

pkg/helpers/helpers.go
```diff
@@ -593,3 +593,18 @@ func UpdateKlusterletRelatedResourcesFn(relatedResources ...operatorapiv1.Relate
 		return nil
 	}
 }
+
+// KlusterletNamespace returns the klusterletNamespace to deploy the agents.
+func KlusterletNamespace(mode operatorapiv1.InstallMode, klusterletName, specNamespace string) string {
+	if mode == operatorapiv1.InstallModeDetached {
+		// return fmt.Sprintf("%s-%s", klusterletName, KlusterletDefaultNamespace)
```
qiujian16: rm this line. Add more comment on this function

pkg/helpers/helpers.go
```diff
@@ -593,3 +593,18 @@ func UpdateKlusterletRelatedResourcesFn(relatedResources ...operatorapiv1.Relate
...
+		return klusterletName
+	}
+
+	if specNamespace == "" {
```
qiujian16: if len(specNamespace) == 0 

it is safer

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -382,11 +410,160 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 	return nil
 }
 
-func (n *klusterletController) ensureNamespace(ctx context.Context, klusterletName, namespace string) error {
-	_, err := n.kubeClient.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+// applyStaticFiles applies files using destKubeclient nad destApiExtensionClient.
+// resource status will be saved in the relatedResourcesStatuses, and will save the
+// result to the klusterlet status if there is any error.
+func (n *klusterletController) applyStaticFiles(ctx context.Context, klusterletName string,
+	staticFiles []string, relatedResourcesStatuses *[]operatorapiv1.RelatedResourceMeta,
```
qiujian16: return relatedResourcesStatuses rathe than adding a pointer

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -310,6 +320,24 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		return err
 	}
 
+	if detachedMode {
+		// In Deatched mode, apply role, rolebinding, service account for registration and work to the management cluster.
+		err = n.applyStaticFiles(ctx, klusterletName, detachedStaticResourceFiles, &relatedResources, &config, n.kubeClient,
```
qiujian16: why the apply static files is different from what you did on managed cluster

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -382,11 +410,160 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+			Type: klusterletApplied, Status: metav1.ConditionFalse, Reason: "KlusterletApplyFailed",
+			Message: fmt.Sprintf("Failed to sync image pull secret to namespace %q: %v", namespace, err),
+		}))
+		return err
+	}
+	return nil
+}
+
+// syncSecret forked from resourceapply.SyncSecret, add an addition snippet to support sync secret to another cluster.
+func syncSecret(client, targetClient coreclientv1.SecretsGetter, recorder events.Recorder,
```
qiujian16: put this in the helper. It is better to add ctx as the parameter

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -185,48 +221,38 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+	// For now, whether in Default or Detached mode, the addons will be deployed on the managed cluster.
+	// sync image pull secret from management cluster to managed cluster for addon namespace
+	// TODO(zhujian7): In the future, we may consider deploy addons on the management cluster in Detached mode.
+	addonNamespace := fmt.Sprintf("%s-addon", config.KlusterletNamespace)
+	err = n.syncPullSecret(ctx, n.kubeClient, managedClusterClients.kubeClient, klusterlet.Name, addonNamespace, controllerContext.Recorder())
+	if err != nil {
+		return err
+	}
 
+	if detachedMode {
```
qiujian16: this is confused..do we really need to apply any role/rolebinding in this ns?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -185,48 +221,38 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
 	// TODO remove this when we do not support kube 1.11 any longer
 	if cnt, err := n.kubeVersion.Compare("v1.12.0"); err == nil && cnt < 0 {
+		errs := []error{}
 		resourceResult := resourceapply.ApplyDirectly(
-			resourceapply.NewKubeClientHolder(n.kubeClient),
+			resourceapply.NewKubeClientHolder(managedClusterClients.kubeClient),
 			controllerContext.Recorder(),
 			func(name string) ([]byte, error) {
-				template, err := manifests.Klusterlet111ManifestFiles.ReadFile(name)
+				template, err := manifests.KlusterletManifestFiles.ReadFile(name)
```
qiujian16: why change this line?

qiujian16 (commented):
I might think it is a bit complicated with a lot of detachMode check in it.

My suggestion is we should separate the manifests and operations to
1. those always put on managed cluster
2. those always put on management cluster (it could be still managedcluster)
3. those you have to check the mode

zhiweiyin318:
do we need to add e2e cases for this scenario? like create 2 kind clusters in action and deploy with detach mode?
or add some integration tests?

zhujian7:
> do we need to add e2e cases for this scenario? like create 2 kind clusters in action and deploy with detach mode? or add some integration tests?

Of course, this is what I am doing these days.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: This func was put in the helper. and I'd prefer to add ctx after upgrading `openshift/library-go` to contain https://github.com/openshift/library-go/pull/1121

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: need to create service account in this ns.

zhujian7: need to create service account in this ns

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: I put all klusterlet files in `KlusterletManifestFiles` to avoid duplication of code

zhujian7:
@qiujian16 all comments have been fixed, PTAL.
@zhiweiyin318 tests were added.

And the RBAC part needs a focused reviewing, please.

Makefile
```diff
@@ -38,14 +38,16 @@ operatorsdk_gen_dir:=$(dir $(OPERATOR_SDK))
 OLM_NAMESPACE?=olm
 OLM_VERSION?=0.16.1
 
-KUSTOMIZE?=$(PERMANENT_TMP_GOPATH)/bin/kustomize
+PWD=$(shell pwd)
```
qiujian16: I think we should update makefile in a followup PR.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -127,23 +143,40 @@ type klusterletConfig struct {
...
-		// AgentCore not found, could have been deleted, do nothing.
+		// Klusterlet not found, could have been deleted, do nothing.
 		return nil
 	}
 	if err != nil {
 		return err
 	}
 	klusterlet = klusterlet.DeepCopy()
+
+	detachedMode := klusterlet.Spec.DeployOption.Mode == operatorapiv1.InstallModeDetached
```
qiujian16: can we use `eq` in the template so we do not need to set this bool var?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -152,10 +185,24 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		ExternalServerURL:         getServersFromKlusterlet(klusterlet),
 		OperatorNamespace:         n.operatorNamespace,
 		Replica:                   helpers.DetermineReplicaByNodes(ctx, n.kubeClient),
+
+		ExternalManagedKubeConfigSecret:             helpers.ExternalManagedKubeConfig,
+		ExternalManagedKubeConfigRegistrationSecret: helpers.ExternalManagedKubeConfigRegistration,
+		ExternalManagedKubeConfigWorkSecret:         helpers.ExternalManagedKubeConfigWork,
+		DetachedMode:                                detachedMode,
```
qiujian16: InstallMode

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -152,10 +185,24 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+		appliedManifestWorkClient: n.appliedManifestWorkClient,
 	}
-	// If namespace is not set, use the default namespace
-	if config.KlusterletNamespace == "" {
-		config.KlusterletNamespace = helpers.KlusterletDefaultNamespace
+
+	if detachedMode {
+		managedClusterClients, err = n.buildManagedClusterClientsDetachedMode(n.kubeClient, config.KlusterletNamespace, config.ExternalManagedKubeConfigSecret)
+		if err != nil {
+			return err
```
qiujian16, line 223: Add a TODO here, you would need to set some status condition as the followup

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -185,104 +232,121 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+	addonNamespace := fmt.Sprintf("%s-addon", config.KlusterletNamespace)
+	err = n.syncPullSecret(ctx, n.kubeClient, managedClusterClients.kubeClient, klusterlet.Name, addonNamespace, controllerContext.Recorder())
+	if err != nil {
+		return err
+	}
 
+	if detachedMode {
+		// In detached mode, we should ensure the namespace on the managed cluster since
+		// some resources(eg:service account) are still deployed on managed cluster.
+		err := n.ensureNamespace(ctx, managedClusterClients.kubeClient, klusterletName, config.KlusterletNamespace)
```
qiujian16, line 269: should we still need this? OR why we only need to ensure this on detached mode?

qiujian16 (commented):
Thanks, I think it looks pretty neat. I still have some comments, but most of the can be done as a followup.

/approve

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: fixed

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: > why we only need to ensure this on detached mode?

In detached mode, the managed cluster and management cluster are separated, the namespace in the management cluster has been [ensured](https://github.com/open-cluster-management-io/registration-operator/pull/172/files#diff-276cfcd05be4cfe9f9381e2a2c7a64e51b09ee5e1c692e3121e7158347c6a166R206) previously.

zhujian7:
@qiujian16 all comments have been fixed, and I add another [commit](https://github.com/open-cluster-management-io/registration-operator/pull/172/commits/591c1fa98c30714aac38ebd4d0f11046270d4ab7)(last commit of this PR) to distinguish the RBAC resource names of managed cluster and management cluster, the RBAC resources names are not in conflict now but may conflict in the future, I think it is necessary to separate. PTAL. Thanks.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: fixed

deploy/klusterlet/olm-catalog/klusterlet/manifests/klusterlet.clusterserviceversion.yaml
```diff
@@ -12,6 +12,30 @@ metadata:
...
+            "registrationImagePullSpec": "quay.io/open-cluster-management/registration",
+            "workImagePullSpec": "quay.io/open-cluster-management/work"
+          }
+        },
+        {
+          "apiVersion": "operator.open-cluster-management.io/v1",
+          "kind": "Klusterlet",
+          "metadata": {
+            "name": "klusterlet"
+          },
```
zhiweiyin318: do not need detached klusterlet in alm-examples. this example user can apply directly on operatorhub. so we cannot put 2 klusterlet here.

zhujian7:
The makefile and README.md will be updated at a follow-up PR.

zhujian7:
@zhiweiyin318 csv updated. PTAL.

PR #176, opened 2021-12-02
add sa syncer

xuezhaojun:

### DIFF
--- a/pkg/helpers/sa_syncer.go
+++ b/pkg/helpers/sa_syncer.go
@@ -1,0 +1,111 @@
+package helpers
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+// EnsureSAToken get the saToken of target sa and ensure it was rendered as expected.
+// A usage example combined with RenderToKubeconfigSecert would be like the following:
+// ...
+// saName := clusterManagerName+"-registration-controller-sa"
+// secretName := saName+"-kubeconfig"
+// err := helpers.EnsureSAToken(ctx, saName, clustermanagerNamespace, amdinKubeClient,
+// 	helpers.RenderToKubeconfigSecret(secretName, clustermanagerNamespace, AdminKubeconfig, kubeClient, recorder)) // kubeClient used to create secret; normally recorder is from controller object.
+// if err != nil {
+// 	return err
+// }
+// ...
+func EnsureSAToken(ctx context.Context, saName, saNamespace string, client kubernetes.Interface, renderSAToken func([]byte) error) error {
+	// get the service account
+	sa, err := client.CoreV1().ServiceAccounts(saNamespace).Get(ctx, saName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if len(sa.Secrets) == 0 {
+		return fmt.Errorf("token secret for %s not exist yet", saName)
+	}
+
+	prefix := saName
+	if len(prefix) > 63 {
+		prefix = prefix[:37]
+	}
+
+	for _, secret := range sa.Secrets {
+		if strings.HasPrefix(secret.Name, prefix) && strings.Contains(secret.Name, "token") {
+			tokenSecretName := secret.Name
+
+			// get the token secret
+			tokenSecret, err := client.CoreV1().Secrets(saNamespace).Get(ctx, tokenSecretName, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+
+			if tokenSecret.Type != corev1.SecretTypeServiceAccountToken {
+				continue
+			}
+
+			saToken, ok := tokenSecret.Data["token"]
+			if !ok {
+				return fmt.Errorf("no token in data for secret %s", tokenSecretName)
+			}
+
+			return renderSAToken(saToken)
+		}
+	}
+
+	return fmt.Errorf("no token secret for this service account %s", sa.Name)
+}
+
+// RenderToKubeconfigSecret would render saToken to a secret.
+func RenderToKubeconfigSecret(secretName, secretNamespace string, templateKubeconfig *rest.Config, client coreclientv1.SecretsGetter, recorder events.Recorder) func([]byte) error {
+	return func(saToken []byte) error {
+		kubeconfigContent, err := clientcmd.Write(clientcmdapi.Config{
+			Kind:       "Config",
+			APIVersion: "v1",
+			Clusters: map[string]*clientcmdapi.Cluster{
+				"cluster": {
+					Server:                   templateKubeconfig.Host,
+					CertificateAuthorityData: templateKubeconfig.CAData,
+				},
+			},
+			Contexts: map[string]*clientcmdapi.Context{
+				"context": {
+					Cluster:  "cluster",
+					AuthInfo: "user",
+				},
+			},
+			AuthInfos: map[string]*clientcmdapi.AuthInfo{
+				"user": {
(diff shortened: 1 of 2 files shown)


pkg/helpers/sa_syncer.go
```diff
@@ -0,0 +1,67 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+// EnsureKubeocnfigFromSA create a secret of kubeconfig with a service account user.
+// It will first get the SA token from the hosted cluster using hostedClient
+// Then it will create a kubeconfig secret on currentClient using currentClient
+// The secret name of kubeconfig should be saName+"-kubeconfig", and the namespace should be saNamespace
+func EnsureKubeconfigFromSA(ctx context.Context, saName, saNamespace string, templateKubeconfig *rest.Config, renderSAToken func([]byte) error, hostedClient kubernetes.Interface) error {
```
qiujian16: I think it is better to not have templateKubeconfig as the parameter, which could be done in render func

xuezhaojun: A little confused here, if we move templateKubeconfig out of parameters, should this method still be called "EnsureKubeconfigFromSA" or just "EnsureSA"? Because users can render it in many ways not just kubeconfig.

xuezhaojun: How about providing two functions for this?

The first function is:

```
func GetSAToken(sa, sanamespace string, hostedClient kubernetes.Interface) ([]byte, error) {...}
``` 

The second function is:
```
func RenderKubeconfig(server, caData, saToken []byte) ([]byte, error) {...}
```
And the client code is:

```
...
saToken,err := helpers.GetSAToken(sa, saNamespace, hostedClient)
if err != nil {
  return err
}
saKubeconfig, err := helpers.RenderKubeconfig(adminKubeconfig.Host, adminKubeconfig. CAData, saToken)
if err != nil {
  return err
}
resourceApply.ApplyResources(...) // apply a secret object
...
```

pkg/helpers/sa_syncer.go
```diff
@@ -0,0 +1,92 @@
...
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+// EnsureSAToken get the saToken of target sa and ensure it was rendered as expected.
+// A usage example combined with RenderToKubeconfigSecert would be like the following:
+// ...
+// saName := clusterManagerName+"-registration-controller-sa"
+// secretName := saName+"-kubeconfig"
+// err := helpers.EnsureSAToken(ctx, cluster, saName, clustermanagerNamespace, hostedAmdinKubeClient,
+// 	helpers.RenderToKubeconfigSecret(secretName, clustermanagerNamespace, hostedAdminKubeconfig, n.kubeClient, n.recorder))
```
zhujian7: Could we change the `n.kubeClient` and `n.recorder` since I don't know what `n` is?

pkg/helpers/sa_syncer.go
```diff
@@ -0,0 +1,92 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+// EnsureSAToken get the saToken of target sa and ensure it was rendered as expected.
+// A usage example combined with RenderToKubeconfigSecert would be like the following:
+// ...
+// saName := clusterManagerName+"-registration-controller-sa"
+// secretName := saName+"-kubeconfig"
+// err := helpers.EnsureSAToken(ctx, cluster, saName, clustermanagerNamespace, hostedAmdinKubeClient,
```
zhujian7: Please check the second param `cluster`, I think it is not needed.

zhujian7 (changes requested):

pkg/helpers/sa_syncer_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	testclient "k8s.io/client-go/kubernetes/fake"
+)
+
+func TestEnsureKubeconfigFromSA(t *testing.T) {
```
zhujian7: Please refresh the test func name, and it's better to add a case to test `RenderToKubeconfigSecret`.

qiujian16 (commented):
/approve

leave this to @zhujian7 to tag

pkg/helpers/sa_syncer.go
xuezhaojun: Done

pkg/helpers/sa_syncer.go
xuezhaojun: Done

pkg/helpers/sa_syncer_test.go
xuezhaojun: Name refreshed

xuezhaojun: And Unit test added.

pkg/helpers/sa_syncer.go
```diff
@@ -0,0 +1,92 @@
...
+func EnsureSAToken(ctx context.Context, saName, saNamespace string, hostedClient kubernetes.Interface, renderSAToken func([]byte) error) error {
+	// get the service account
+	sa, err := hostedClient.CoreV1().ServiceAccounts(saNamespace).Get(ctx, saName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if len(sa.Secrets) == 0 {
+		return fmt.Errorf("token secret for %s not exist yet", saName)
+	}
+	tokenSecretName := sa.Secrets[0].Name
```
zhiweiyin318: do we need to check the name of secret in sa, like xxx-token-xx?

xuezhaojun: I thought the token secret would be the first secret as default, I don't know whether it was promised or not. It's based on observation. 
However, I'm also not sure about the secret name. Does the format of the name **have to** be "xxx-token-xx"? Is there any case that the token secret's name doesn't contain a "token"?

xuezhaojun: Done. Now if no secret's name contains "token", the function would return an error.

zhujian7:
I'd suggest changing `hostedxxx` to `managementxxx` or something else that does not contain `hosted`.

xuezhaojun:
> I'd suggest changing `hostedxxx` to `managementxxx` or something else that does not contain `hosted`.

Good point! Code Updated.

zhujian7:
LGTM, leave @zhiweiyin318  to tag

pkg/helpers/sa_syncer.go
```diff
@@ -0,0 +1,99 @@
...
+	sa, err := client.CoreV1().ServiceAccounts(saNamespace).Get(ctx, saName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if len(sa.Secrets) == 0 {
+		return fmt.Errorf("token secret for %s not exist yet", saName)
+	}
+	for _, secret := range sa.Secrets {
+		if strings.Contains(secret.Name, "token") {
+			tokenSecretName := sa.Secrets[0].Name
```
zhiweiyin318: alway get the name of as.Secrets[0] here in loop.

pkg/helpers/sa_syncer.go
```diff
@@ -0,0 +1,99 @@
...
+	// get the service account
+	sa, err := client.CoreV1().ServiceAccounts(saNamespace).Get(ctx, saName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if len(sa.Secrets) == 0 {
+		return fmt.Errorf("token secret for %s not exist yet", saName)
+	}
+	for _, secret := range sa.Secrets {
+		if strings.Contains(secret.Name, "token") {
```
zhiweiyin318: I checked that the 0 index of secret in sa is not always saName-token-xxx, it could be saName-dockercfg-xx.
what if there is a "token" string in the sa name?
I think you can ref this https://github.com/open-cluster-management/managedcluster-import-controller/blob/884905557ba9e956ec7f975c01c72a7e6d577307/pkg/controller/importconfig/cluster_info.go#L37 
if you want just check secret name, could consider using `<saName>-token- `

xuezhaojun: Thanks for the feedback, it's a great reference. Now we set triple checks to make sure we get the correct result.
1. prefix check
2. token in name check
3. secret type check

pkg/helpers/sa_syncer.go
xuezhaojun: Solved.

Issue #157, opened 2021-10-28
Allow cluster manager to run outside of hub cluster


qiujian16:
We should be able to run cluster manager and all related runtimes outside of the hub cluster, but in a different meta management cluster.

This allows us to create multiple hub clusters on meta managed cluster

/assign @xuezhaojun

qiujian16:
this is finished

/close

PR #178, opened 2021-12-17
Fix apiservice reapply infinitely issue

xuezhaojun:

Relevant issue: https://github.com/open-cluster-management-io/registration-operator/issues/157

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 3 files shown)


pkg/helpers/helpers.go
```diff
@@ -334,10 +335,16 @@ func ApplyDirectly(
 			result.Result, result.Changed, result.Error = ApplyMutatingWebhookConfiguration(
 				client.AdmissionregistrationV1(), t)
 		case *apiregistrationv1.APIService:
+			t.ObjectMeta.Annotations = make(map[string]string)
+			checksum := md5.Sum(t.Spec.CABundle)
+			t.ObjectMeta.Annotations["caBundle-checksum"] = string(checksum[:]) // to trigger the update when caBundle changed
```
qiujian16, line 340: I think we commonly use sha256 like this

```
fmt.Sprintf("%x", sha256.Sum256(data))
```

xuezhaojun: Yes, changed to use sha56.

PR #180, opened 2022-01-06
update README for running klusterlet in detached mode

zhujian7:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,8 @@ operatorsdk_gen_dir:=$(dir $(OPERATOR_SDK))
 OLM_NAMESPACE?=olm
 OLM_VERSION?=0.16.1
 
-KUSTOMIZE?=$(PERMANENT_TMP_GOPATH)/bin/kustomize
+PWD=$(shell pwd)
+KUSTOMIZE?=$(PWD)/$(PERMANENT_TMP_GOPATH)/bin/kustomize
 KUSTOMIZE_VERSION?=v3.5.4
 KUSTOMIZE_ARCHIVE_NAME?=kustomize_$(KUSTOMIZE_VERSION)_$(GOHOSTOS)_$(GOHOSTARCH).tar.gz
 kustomize_dir:=$(dir $(KUSTOMIZE))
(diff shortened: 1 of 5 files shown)


deploy/klusterlet/config/samples/managedcluster/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+secretGenerator:
+  - name: external-managed-kubeconfig
+    namespace: klusterlet
+    files:
+      - kubeconfig=external-managed-kubeconfig
+    type: "Opaque"
+
+generatorOptions:
+  disableNameSuffixHash: true
```
xuezhaojun: need a newline.

Makefile
```diff
@@ -100,6 +107,8 @@ clean-hub: clean-hub-cr clean-hub-operator
 
 clean-spoke: clean-spoke-cr clean-spoke-operator
 
+clean-spoke-detached: clean-spoke-cr-detached clean-spoke-operator
```
xuezhaojun, line 110: Since now we can deploy several spoke on one cluster, I wonder whether do we have the case that the user only wants to clean a specific spoke on the cluster?

deploy/klusterlet/olm-catalog/klusterlet/manifests/klusterlet.clusterserviceversion.yaml
```diff
@@ -209,6 +209,13 @@ spec:
           verbs:
           - update
           - patch
+        - apiGroups:
+          - work.open-cluster-management.io
```
zhiweiyin318: need update rbac in config dir firstly, and then update csv by `make update-csv`

Makefile
zhujian7: Do you mean a user only wants to clean the klusterlet for a specific spoke cluster? I think deleting the `Klusterlet` directly can help.
The target `clean-spoke-detached` in Makefile just covers the case to clean spoke CRs and operator that the `deploy-spoke-detached` target created.

xuezhaojun: Yes, that makes sense.

PR #181, opened 2022-01-07
get the namespace of klusterlet from helper

zhiweiyin318:
refactor the helper.KlusterletNamespace() 
fix the issue that HubConfigSecretDegraded  condation cannot be  recovered in detach mode since the ns of klusterlet is incorrect. 

### DIFF
--- a/pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
+++ b/pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
@@ -44,7 +44,6 @@ type klusterletLocker struct {
 }
 
 const (
-	klusterletNamespace     = "open-cluster-management-agent"
 	bootstrapSecret         = "BootstrapSecret"
 	bootstrapSecretDegraded = "BootstrapSecretDegraded"
 	hubConfigSecret         = "HubConfigSecret"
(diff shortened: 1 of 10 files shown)


pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
```diff
@@ -123,10 +123,7 @@ func (c *ssarController) sync(ctx context.Context, controllerContext factory.Syn
 		defer c.deleteSSARChecking(klusterletName)
 
 		klog.V(4).Infof("Reconciling Klusterlet %q", klusterletName)
-		klusterletNS := klusterlet.Spec.Namespace
-		if klusterletNS == "" {
-			klusterletNS = klusterletNamespace
-		}
+		klusterletNS := helpers.KlusterletNamespace(klusterlet)
```
zhujian7, line 125: I think we can delete the const `klusterletNamespace` since it will not be used.

zhiweiyin318: sure. 
@qiujian16  could we put the const vars (like the condition names, secret names, namesapce names etc.)  of the packages into a specified file?  some const may be defined repeatedly if we use them in multi packages.

qiujian16: yes, I think we should maintain the condition types in APIs. secret/namespace depends. A namespace defined in API may also should be defined as a const in API.

qiujian16 (commented):
/approve
/lgtm

for ns/condition type, would you create another PR as followup?

PR #184, opened 2022-01-10
scale up the work agent deployment after the hub-kuebeconfig-secret is created

zhiweiyin318:
fix issue https://github.com/open-cluster-management-io/community/issues/111

scale up the work agent deployment when the hubConnectionDegraded condition of klusterlet is False. 

### DIFF
--- a/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
+++ b/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
@@ -43,6 +43,7 @@ const (
 	imagePullSecret              = "open-cluster-management-image-pull-credentials"
 	klusterletApplied            = "Applied"
 	klusterletReadyToApply       = "ReadyToApply"
+	hubConnectionDegraded        = "HubConnectionDegraded"
 	appliedManifestWorkFinalizer = "cluster.open-cluster-management.io/applied-manifest-work-cleanup"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -335,15 +330,22 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
 	relatedResources = append(relatedResources, statuses...)
 
-	// If cluster name is empty, read cluster name from hub config secret
+	// If cluster name is empty, read cluster name from hub config secret.
+	// registration-agent generated the cluster name and set it into hub config secret.
 	if config.ClusterName == "" {
-		clusterName := hubSecret.Data["cluster-name"]
-		if clusterName != nil {
-			config.ClusterName = string(clusterName)
+		config.ClusterName, err = n.getClusterNameFromHubKubeConfigSecret(ctx, &config)
```
qiujian16: if using pointer here. it does not need to return and set, right?

zhiweiyin318: no, I don't think we should deploy work agent deployment if clusterName is empty when failed to get hub-kubeconfig secret or there is no clusterName in hub-kubeconfig secret.

zhiweiyin318: I got your point.  you mean this ?
```
clusterName := n.getClusterNameFromHubKubeConfigSecret(ctx, &config)
if clusterName == nil {
 return fmt.Errorf("failed to get clusterName")
}
```

qiujian16:
```go
err := n.setClusterNameFromHubKubeConfigSecret(ctx, &config)
if err != nil {
  return err
}
```

zhiweiyin318: done

yue9944882 (commented):
LGTM

PR #191, opened 2022-01-14
add nil check for nil clients in helper

xuezhaojun:

**Issue:**
When we want to use `ApplyDirecly` for the management cluster, we don't have apiExteonsionClient, apiRegistrationClient, because the management cluster has **no CRD and APIService** resources on it.
 
**Value:**
This PR will allow us set those two arguments to `nil` when using the `ApplyDirectly` func .

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/onsi/gomega v1.10.1
 	github.com/openshift/api v0.0.0-20210331193751-3acddb19d360
 	github.com/openshift/build-machinery-go v0.0.0-20211213093930-7e33a7eb4ce3
-	github.com/openshift/library-go v0.0.0-20210406144447-d9cdfbd844ea
+	github.com/openshift/library-go v0.0.0-20210407110438-80b76d711afb
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	k8s.io/api v0.21.1
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -193,11 +193,23 @@ func CleanUpStaticObject(
 	case *rbacv1.RoleBinding:
 		err = client.RbacV1().RoleBindings(t.Namespace).Delete(ctx, t.Name, metav1.DeleteOptions{})
 	case *apiextensionsv1.CustomResourceDefinition:
-		err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().Delete(ctx, t.Name, metav1.DeleteOptions{})
+		if apiExtensionClient == nil {
+			err = fmt.Errorf("apiExtensionClient is nil")
+		} else {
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/onsi/gomega v1.10.1
 	github.com/openshift/api v0.0.0-20210331193751-3acddb19d360
 	github.com/openshift/build-machinery-go v0.0.0-20211213093930-7e33a7eb4ce3
-	github.com/openshift/library-go v0.0.0-20210406144447-d9cdfbd844ea
+	github.com/openshift/library-go v0.0.0-20210401181443-5cf97eb29077
```
qiujian16: why we use a lower version of library-go?

xuezhaojun: Fixed but it still confused me too: 

First the bug was fixed by [5cf97eb29077b69d28be0b350260d8c1f40a64d4](https://github.com/openshift/library-go/blob/5cf97eb29077b69d28be0b350260d8c1f40a64d4/pkg/operator/resource/resourceapply/generic.go) in 2 Apr.

But after that, the modification disappeared until [80b76d711afbe5d39d8240da7a539f567bcdcf05](https://github.com/openshift/library-go/blob/80b76d711afbe5d39d8240da7a539f567bcdcf05/pkg/operator/resource/resourceapply/generic.go) 7 Apr

So commits between 04.02 to 04.07, the bug still existed.

---

I have changed lib-go to `80b76d711afbe5d39d8240da7a539f567bcdcf05` which is `20210407110438`.

pkg/helpers/helpers.go
```diff
@@ -193,11 +193,23 @@ func CleanUpStaticObject(
 	case *rbacv1.RoleBinding:
 		err = client.RbacV1().RoleBindings(t.Namespace).Delete(ctx, t.Name, metav1.DeleteOptions{})
 	case *apiextensionsv1.CustomResourceDefinition:
-		err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().Delete(ctx, t.Name, metav1.DeleteOptions{})
```
zhiweiyin318, line 196: which case where the `apiExtensionClient` is set to be nil? Is this also possible for client?
I think it's better to check them at the beginning in the func, then we don't need to check in each case.

pkg/helpers/helpers.go
```diff
@@ -193,11 +193,23 @@ func CleanUpStaticObject(
...
+		if apiExtensionClient == nil {
+			err = fmt.Errorf("apiExtensionClient is nil")
```
zhiweiyin318, line 197: just a confusion here. all the places where we call` helpers.ApplyDirectly` and` helpers.CleanUpStaticObject `have validate apiExtensionClient and apiRegistrationClient. 
which scenario the imput clients are nil?

zhiweiyin318: ignore this. I missed that  management cluster has no CRD and APIService resources.

Issue #196, opened 2022-01-22
SA token ensure func did not find the right tokens


zhujian7:
When the service-account name is long enough, the [sa token ensure func](https://github.com/open-cluster-management-io/registration-operator/blob/v0.6.0/pkg/helpers/sa_syncer.go#L47) may mismatch the token.

For example, token names of the following service account do not have "token" characters.
```
apiVersion: v1
imagePullSecrets:
- name: open-cluster-management-image-pull-credentials
- name: klusterlet-hypershift-demo-managed-clu-f203af0b-dockercfg-kgh65
kind: ServiceAccount
metadata:
  creationTimestamp: "2022-01-22T07:56:39Z"
  name: klusterlet-hypershift-demo-managed-cluster-registration-sa
  namespace: klusterlet-hypershift-demo-managed-cluster
  resourceVersion: "70692537"
  uid: 975eb940-6dbd-4ad7-a2d8-77cfd867887f
secrets:
- name: klusterlet-hypershift-demo-managed-cluster-registration-sax7zps
- name: klusterlet-hypershift-demo-managed-clu-f203af0b-dockercfg-kgh65
```

cc @xuezhaojun

xuezhaojun:
That's a good point! After a thinking, how about change the code from

```go
for _, secret := range sa.Secrets {
		if strings.HasPrefix(secret.Name, prefix) && strings.Contains(secret.Name, "token") {
			tokenSecretName := secret.Name

			// get the token secret
			tokenSecret, err := saClient.CoreV1().Secrets(saNamespace).Get(ctx, tokenSecretName, metav1.GetOptions{})
			if err != nil {
				return err
			}

			if tokenSecret.Type != corev1.SecretTypeServiceAccountToken {
				continue
			}

			saToken, ok := tokenSecret.Data["token"]
			if !ok {
				return fmt.Errorf("no token in data for secret %s", tokenSecretName)
			}

			return renderSAToken(saToken)
		}
}
```

 to:

```go
for _, secret := range sa.Secrets {
		tokenSecretName := secret.Name

		// get the token secret
		tokenSecret, err := saClient.CoreV1().Secrets(saNamespace).Get(ctx, tokenSecretName, metav1.GetOptions{})
		if err != nil {
			return err
		}

		if tokenSecret.Type != corev1.SecretTypeServiceAccountToken {
			continue
		}

		saToken, ok := tokenSecret.Data["token"]
		if !ok {
[...]

PR #197, opened 2022-01-22
Fixed can't get sa token secret when secret name is long

xuezhaojun:

Relevant issue: https://github.com/open-cluster-management-io/registration-operator/issues/196


xuezhaojun:
@zhiweiyin318 Hello, Zhiwei, would you like to have a review?

PR #202, opened 2022-02-10
Refactor on connection check in klusterlet

qiujian16:
1. Update observerved generation
2. Code update to clear the logic


qiujian16:
may need to add more negative test cases.

skeeey:
> may need to add more negative test cases.

yeah, we have some negative test cases in the integration, we can enhance them :)

PR #210, opened 2022-03-03
enable feature gate for registration webhook

ldpliu:
Signed-off-by: ldpliu <daliu@redhat.com>


qiujian16:
@ldpliu would you also check the rbac permission of the registration controller

ldpliu:
> @ldpliu would you also check the rbac permission of the registration controller

Yes, we already have clusterset create permission in both [registration](https://github.com/open-cluster-management-io/registration/blob/b8a1f9d676a3d2aeb68482d96ce53739c3fc867e/deploy/hub/hub_controller_clusterrole.yaml#L53) and [registrition-operator](https://github.com/open-cluster-management-io/registration-operator/blob/c1b1067f54be39473b189b0aa34100814995e198/manifests/cluster-manager/hub/cluster-manager-registration-clusterrole.yaml#L53)

PR #213, opened 2022-03-10
only set work-agent replica 0 when hub-kubeconfig-secret is missing

zhiweiyin318:
when the cluster is deleting, the work-agent needs to continue working until the manifestworks cleanup is finished.

### DIFF
--- a/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
+++ b/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
@@ -44,6 +44,7 @@ const (
 	klusterletApplied            = "Applied"
 	klusterletReadyToApply       = "ReadyToApply"
 	hubConnectionDegraded        = "HubConnectionDegraded"
+	hubKubeConfigSecretMissing   = "HubKubeConfigSecretMissing"
 	appliedManifestWorkFinalizer = "cluster.open-cluster-management.io/applied-manifest-work-cleanup"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -44,6 +44,7 @@ const (
 	klusterletApplied            = "Applied"
 	klusterletReadyToApply       = "ReadyToApply"
 	hubConnectionDegraded        = "HubConnectionDegraded"
+	hubKubeConfigSecretMissing   = "HubKubeConfigSecretMissing"
```
qiujian16, line 47: should this be a const somewhere that statuscontroller can also use?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -339,12 +340,16 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		}
 	}
 	// Deploy work agent
-	// scale up the work agent deployment when the hubConnectionDegraded condition is False.
+	// scale up the work agent deployment when the hubConnectionDegraded condition is True and hub-kubeconfig-secret is missing.
```
qiujian16: please doc more context why we need to check the reason?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhiweiyin318: I will do it as a followup.  there are several const variables/strings need to be used as a global one.
not sure which package is the proper place, maybe need to create a new package.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhiweiyin318: done

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -339,12 +340,17 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		}
 	}
 	// Deploy work agent
-	// scale up the work agent deployment when the hubConnectionDegraded condition is False.
-	// because the work agent deployment has a dependency of hub-kubeconfig-secret secret.
+	// do not scale up work agent pods if there is no hub-kubeconfig-secret when klusterelt CR is created, because
+	// the work agent deployment has a dependency of hub-kubeconfig-secret secret.
+	// do nod scale down work agent pods if the cluster is deleting since the work-agent needs to do some cleanup works.
```
qiujian16: I would say like

work agent is scaled to 0 only when degrade is true withe reason if configmissing. It is to ensure a fast startup of work agent when the klusterlet is bootstrapped at the first time. The work agent should not be scaled to 0 in degraded condition with other reasons, because we still need work agent running even though the hub kubconfig is missing some certain permission. It can ensure work agent to clean up the resources defined in manifestworks when cluster is unjoining from the hub.

zhiweiyin318: modified.

PR #220, opened 2022-03-15
Add hosted mode in klusterlet

xuezhaojun:

### DIFF
--- a/manifests/klusterlet/management/klusterlet-registration-deployment.yaml
+++ b/manifests/klusterlet/management/klusterlet-registration-deployment.yaml
@@ -51,7 +51,7 @@ spec:
           {{if .ExternalServerURL}}
           - "--spoke-external-server-urls={{ .ExternalServerURL }}"
           {{end}}
-          {{if eq .InstallMode "Detached"}}
+          {{if eq .InstallMode "Hosted"}}
           - "--spoke-kubeconfig=/spoke/config/kubeconfig"
           {{end}}
           {{if eq .Replica 1}}
@@ -70,7 +70,7 @@ spec:
           readOnly: true
         - name: hub-kubeconfig
           mountPath: "/spoke/hub-kubeconfig"
-        {{if eq .InstallMode "Detached"}}
+        {{if eq .InstallMode "Hosted"}}
         - name: spoke-kubeconfig-secret
           mountPath: "/spoke/config"
           readOnly: true
@@ -99,7 +99,7 @@ spec:
       - name: hub-kubeconfig
         emptyDir:
           medium: Memory
-      {{if eq .InstallMode "Detached"}}
+      {{if eq .InstallMode "Hosted"}}
       - name: spoke-kubeconfig-secret
         secret:
           secretName: {{ .ExternalManagedKubeConfigRegistrationSecret }}
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -600,7 +600,7 @@ func LoadClientConfigFromSecret(secret *corev1.Secret) (*rest.Config, error) {
 func DetermineReplica(ctx context.Context, kubeClient kubernetes.Interface, mode operatorapiv1.InstallMode) int32 {
 	// For hosted mode, there may be many cluster-manager/klusterlet running on the management cluster,
 	// set the replica to 1 to reduce the footprint of the management cluster.
-	if mode == operatorapiv1.InstallModeDetached {
+	if mode == operatorapiv1.InstallModeDetached || mode == operatorapiv1.InstallModeHosted {
 		return singleReplica
 	}
 	return DetermineReplicaByNodes(ctx, kubeClient)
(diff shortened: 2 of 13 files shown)


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -209,8 +209,8 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		appliedManifestWorkClient: n.appliedManifestWorkClient,
 	}
 
-	if config.InstallMode == operatorapiv1.InstallModeDetached {
-		managedClusterClients, err = n.buildManagedClusterClientsDetachedMode(ctx, n.kubeClient, config.KlusterletNamespace, config.ExternalManagedKubeConfigSecret)
+	if config.InstallMode == operatorapiv1.InstallModeDetached || config.InstallMode == operatorapiv1.InstallModeHosted {
```
zhujian7: I am thinking can we just change the installMode value to `Hosted` when it is `Detached` here(and add a comment that we will depredate the `Detached`), so we can remove `Detached` at other places？ The advantage of this is to cancel Detached in the future just need to change here.

zhujian7:
Great job! looks good to me.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
xuezhaojun: That's a good point!  Updated, now we replace the mode to hosted at the first place if the mode is detached.

manifests/klusterlet/management/klusterlet-registration-deployment.yaml
```diff
@@ -70,7 +70,7 @@ spec:
           readOnly: true
         - name: hub-kubeconfig
           mountPath: "/spoke/hub-kubeconfig"
-        {{if eq .InstallMode "Detached"}}
+        {{if or (eq .InstallMode "Detached") (eq .InstallMode "Hosted")}}
```
qiujian16: you would not need this or since the mode is always hosted, right?

pkg/helpers/helpers.go
```diff
@@ -600,7 +600,7 @@ func LoadClientConfigFromSecret(secret *corev1.Secret) (*rest.Config, error) {
 func DetermineReplica(ctx context.Context, kubeClient kubernetes.Interface, mode operatorapiv1.InstallMode) int32 {
 	// For hosted mode, there may be many cluster-manager/klusterlet running on the management cluster,
 	// set the replica to 1 to reduce the footprint of the management cluster.
-	if mode == operatorapiv1.InstallModeDetached {
+	if mode == operatorapiv1.InstallModeDetached || mode == operatorapiv1.InstallModeHosted {
```
qiujian16, line 603: ditto

manifests/klusterlet/management/klusterlet-registration-deployment.yaml
```diff
@@ -51,7 +51,7 @@ spec:
           {{if .ExternalServerURL}}
           - "--spoke-external-server-urls={{ .ExternalServerURL }}"
           {{end}}
-          {{if eq .InstallMode "Detached"}}
+          {{if or (eq .InstallMode "Detached") (eq .InstallMode "Hosted")}}
```
zhujian7: I think this `Detached` can be removed too.

pkg/helpers/helpers.go
xuezhaojun: That sounds right! fixed.

manifests/klusterlet/management/klusterlet-registration-deployment.yaml
xuezhaojun: Fixed.

manifests/klusterlet/management/klusterlet-registration-deployment.yaml
xuezhaojun: Fixed.

pkg/helpers/helpers.go
```diff
@@ -736,9 +736,9 @@ func UpdateKlusterletRelatedResourcesFn(relatedResources ...operatorapiv1.Relate
 }
 
 // KlusterletNamespace returns the klusterletNamespace to deploy the agents.
-// Note in Detached mode, the specNamespace will be ignored.
+// Note in Hosted mode, the specNamespace will be ignored.
 func KlusterletNamespace(klusterlet *operatorapiv1.Klusterlet) string {
-	if klusterlet.Spec.DeployOption.Mode == operatorapiv1.InstallModeDetached {
+	if klusterlet.Spec.DeployOption.Mode == operatorapiv1.InstallModeHosted {
```
qiujian16: hrm, this would still need consider detached?

pkg/helpers/helpers.go
```diff
@@ -600,7 +600,7 @@ func LoadClientConfigFromSecret(secret *corev1.Secret) (*rest.Config, error) {
 func DetermineReplica(ctx context.Context, kubeClient kubernetes.Interface, mode operatorapiv1.InstallMode) int32 {
 	// For hosted mode, there may be many cluster-manager/klusterlet running on the management cluster,
 	// set the replica to 1 to reduce the footprint of the management cluster.
-	if mode == operatorapiv1.InstallModeDetached {
+	if mode == operatorapiv1.InstallModeHosted {
```
qiujian16: it is fine if it is read from config, but detached needs to be considered if it is read from klusterlet resource

pkg/helpers/helpers.go
xuezhaojun: I'm not sure about the "read from klusterlet resource" case, will this case have affect on this function tool?

pkg/helpers/helpers.go
xuezhaojun: Sure, updated.

pkg/helpers/helpers.go
xuezhaojun: Updated  with Detached also considered.

qiujian16 (commented):
/approve

PR #227, opened 2022-03-29
support Klusterlet.Spec.Namespace in hosted mode

elgnay:

### DIFF
--- a/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
+++ b/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
@@ -143,8 +143,19 @@ func NewKlusterletController(
 
 // klusterletConfig is used to render the template of hub manifests
 type klusterletConfig struct {
-	KlusterletName            string
-	KlusterletNamespace       string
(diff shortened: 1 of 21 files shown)


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -143,8 +143,15 @@ func NewKlusterletController(
 
 // klusterletConfig is used to render the template of hub manifests
 type klusterletConfig struct {
-	KlusterletName            string
-	KlusterletNamespace       string
+	KlusterletName string
+	// KlusterletNamespace is the klusterlet namespace on the managed cluster.
```
zhujian7: Can we add more comments on this to explain something like:
- in default mode, it is identical to the agent namespace.
- in hosted mode, we store some configuration in this namespace that will be used by the agents on the remote management cluster.

elgnay: Doc updated for those two fields.

qiujian16 (commented):
/approve

PR #228, opened 2022-03-30
upgrade lib

zhiweiyin318:


go.mod
```diff
@@ -4,24 +4,24 @@ go 1.17
...
 	github.com/onsi/ginkgo v1.16.5
-	github.com/onsi/gomega v1.17.0
-	github.com/openshift/api v0.0.0-20211209135129-c58d9f695577
-	github.com/openshift/build-machinery-go v0.0.0-20211213093930-7e33a7eb4ce3
-	github.com/openshift/library-go v0.0.0-20220308152156-227dd2b19774
-	github.com/spf13/cobra v1.2.1
+	github.com/onsi/gomega v1.19.0
+	github.com/openshift/api v0.0.0-20220329125147-1a6fa2913810
+	github.com/openshift/build-machinery-go v0.0.0-20220121085309-f94edc2d6874
+	github.com/openshift/library-go v0.0.0-20220329193146-715792ed530d
```
elgnay, line 12: Upgrading library-go to the latest version may result in leader election issue on OCP311. Maybe we can disable the leader election of klusterlet-operator since its replica is 1 by default. @qiujian16 What do you think?

elgnay, line 12: Upgrading library-go to the latest version may result in leader election issue on OCP311. Maybe we can disable the leader election of klusterlet-operator since its replica is 1 by default. @qiujian16 What do you think?

qiujian16: we already did that in import controller when I upgrade the library last time :)

PR #253, opened 2022-06-27
Add controller to sync image pull secret into addon namespaces.

xuezhaojun:

Epic: 
https://github.com/stolostron/backlog/issues/23093

User-story:
https://github.com/stolostron/backlog/issues/23929


pkg/operators/klusterlet/controllers/addoncontroller/addoncontroller.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func NewAddonPullImageSecretController(kubeClient kubernetes.Interface, operatorNamespace string, namespaceInformer coreinformer.NamespaceInformer, recorder events.Recorder) factory.Controller {
+	ac := &addonPullImageSecretController{
+		operatorNamespace: operatorNamespace,
+		namespaceInformer: namespaceInformer,
+		kubeClient:        kubeClient,
+		recorder:          recorder,
+	}
+	return factory.New().WithInformersQueueKeyFunc(func(o runtime.Object) string {
```
qiujian16: add a filter here.

pkg/operators/klusterlet/controllers/addoncontroller/addoncontroller.go
```diff
@@ -0,0 +1,78 @@
...
+		c.operatorNamespace,
+		imagePullSecret,
+		namespace,
+		imagePullSecret,
+		[]metav1.OwnerReference{},
+	)
+	if err != nil {
+		return err
+	}
+	return nil
```
qiujian16: do we need to remove the secret when annotation is removed?

qiujian16 (commented):
which kubeclient is used here?

zhiweiyin318:
we could remove the sync pull secret part for the open-cluster-management-agent-addon ns if we add the annotation for the open-cluster-management-agent-addon ns.

pkg/operators/klusterlet/controllers/addoncontroller/addoncontroller.go
xuezhaojun: filter added.

pkg/operators/klusterlet/controllers/addoncontroller/addoncontroller.go
xuezhaojun: Currently, we should keep the secret because **users could also create a namespace with the secret**. 
So `(a namespace) + (With image pull secret) + (Without the label)` is a **legal** situation.

xuezhaojun:
> which kubeclient is used here?

Should be the InCluster Kubeclient.

qiujian16 (commented):
you might want to remove the sync secret code in klusterlet controller, and add label for open-cluster-management-xxx ns

xuezhaojun:
> you might want to remove the sync secret code in klusterlet controller, and add label for open-cluster-management-xxx ns

We have the discussion this morning with @elgnay  @zhiweiyin318 and decide to keep it.

Because the klusterlet-controller not only syncs the secret to **management cluster's `open-cluster-management-agent-addon` ns** but also syncs the secret to **the managed cluster's `open-cluster-management-agent-addon` ns**.

https://github.com/open-cluster-management-io/registration-operator/blob/a7dd402db878b3e40bfb95d1d35582d129250ea0/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go#L341

As the above code shows, we are using `managedClusterClients.kubeClient`.

So if we remove sync secret from  klusterlet-controller, we now only sync secret to managementCluster but not sync secret to managed cluster.

case:

If any addons is now installed in `open-cluster-managemetn-agent-addon` ns in the managed-cluster (meanwhile the klusterlet is running in Hosted Mode on management cluster), these addons will be affect.

The addons **will not create** ns by themselves because **addons are in default mode**.

pkg/operators/klusterlet/controllers/addoncontroller/addoncontroller.go
xuezhaojun: Agreed to keep secret. @zhiweiyin318 @elgnay

qiujian16:
ok, it needs to be doc'ed in the code.

xuezhaojun:
> ok, it needs to be doc'ed in the code.

The reason is doced, PTAL @qiujian16

PR #255, opened 2022-06-28
keep appliedmanifestwork & clusterclaim CRDs when unstalling klusterlet

elgnay:

### DIFF
--- a/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
+++ b/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
@@ -656,7 +656,7 @@ func (n *klusterletController) cleanUp(
 		return err
 	}
 
-	// remove AppliedManifestWorks, should be executed **before** "remove the CRDs" and "remove hub kubeconfig secret"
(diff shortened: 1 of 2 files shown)


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -712,20 +712,6 @@ func (n *klusterletController) cleanUp(
 		}
 	}
 
-	// remove the CRDs
```
qiujian16, line 715: do we need to remove the CR?

elgnay: As we discussed in slack, it's fine to leave the CRs for now.

qiujian16: please add a comment here indicating we do not delete crd and CR and the reason.

qiujian16 (commented):
/approve
/lgtm

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
elgnay: Comment added.

PR #259, opened 2022-07-04
Use token request to sync sa

qiujian16:

### DIFF
--- a/pkg/helpers/sa_syncer.go
+++ b/pkg/helpers/sa_syncer.go
@@ -2,122 +2,183 @@ package helpers
 
 import (
 	"context"
-	"fmt"
 	"io/ioutil"
+	"time"
 
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	authv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


pkg/helpers/sa_syncer.go
```diff
@@ -16,108 +17,140 @@ import (
...
+			}
+
+			return saToken, nil
 		}
 
-		saToken, ok := tokenSecret.Data["token"]
-		if !ok {
-			return fmt.Errorf("no token in data for secret %s", tokenSecretName)
+		// 8640 hour
+		expirationSec := int64(8640 * 3600)
```
zhujian7: what will happen after this period? will the klusterlet refresh the token?

xuezhaojun:
LGTM

pkg/helpers/sa_syncer.go
qiujian16: I add a TODO to check the validity of the token and refresh if necessary

qiujian16:
/hold

still need to add rbac setting.

pkg/helpers/sa_syncer.go
```diff
@@ -2,122 +2,183 @@ package helpers
...
+		if lifetime < refreshThreshold {
+			return false
+		}
 	}
 
-	return fmt.Errorf("no token secret for this service account %s", sa.Name)
+	return true
 }
 
 // RenderToKubeconfigSecret would render saToken to a secret.
```
zhujian7: // applyKubeconfigSecret ....

PR #262, opened 2022-07-11
Allow OCM addons to setup metrics collection with Prometheus

mprahl:
Related:
https://github.com/stolostron/backlog/issues/23547

### DIFF
--- a/manifests/klusterlet/managed/klusterlet-work-clusterrole-execution.yaml
+++ b/manifests/klusterlet/managed/klusterlet-work-clusterrole-execution.yaml
@@ -20,3 +20,9 @@ rules:
 - apiGroups: ["rbac.authorization.k8s.io"]
   resources: ["clusterroles", "roles"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete", "escalate", "bind"]
+# Allow OCM addons to setup metrics collection with Prometheus
+# TODO: Move this permission to the open-cluster-management:{{ .KlusterletName }}-work:execution Role (not ClusterRole)
+# when it is created.
+- apiGroups: ["monitoring.coreos.com"]
+  resources: ["servicemonitors"]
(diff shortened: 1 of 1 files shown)


manifests/klusterlet/managed/klusterlet-work-clusterrole-execution.yaml
```diff
@@ -20,3 +20,7 @@ rules:
 - apiGroups: ["rbac.authorization.k8s.io"]
   resources: ["clusterroles", "roles"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete", "escalate", "bind"]
+# Allow OCM addons to setup metrics collection with Prometheus
+- apiGroups: ["monitoring.coreos.com"]
+  resources: ["servicemonitors"]
```
haoqing0110, line 27: `servicemonitors` is a namespace scoped CRD, as we discussed before in [Code refactor to split permission for registration/work](https://docs.google.com/document/d/1SWT4jcBhn6c3dPt2TyNdh4NqfdxSSDiOlSSYgy6QgqM/edit#heading=h.ba0yojrlee5r), this namespace scoped resource should be put this into `Role open-cluster-management:{{ .KlusterletName }}-work:execution` (this Role is not created so far). @qiujian16 any comments for this?

qiujian16: I think this is for addon to create their service monitors? It is  ok to put here if we do not have that role yet, but yes we should put it in a role in the future. A TODO here would be helpful.

mprahl: That's right. It's for an addon to create a service monitor.

I just added the `TODO` comment as requested.

haoqing0110 (approved):
/lgtm

mprahl:
> /lgtm

Could you also please add the `/approve` comment?

PR #267, opened 2022-08-01
fix CVE-2021-38561

elgnay:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -115,3 +115,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace golang.org/x/text => golang.org/x/text v0.3.7 // fix CVE-2021-38561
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -115,3 +115,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace golang.org/x/text => golang.org/x/text v0.3.7 // fix CVE-2021-38561
```
qiujian16, line 119: why it is a replace?

elgnay: Without the replace, even we have already specified the version of `golang.org/x/text` to v0.3.7, there are still some previous versions of this lib referenced in the go.sum file.

elgnay: By runing `go list -m golang.org/x/text`, the version on which registration-operator really depends will be shown.
```
acm/registration-operator$ go list -m golang.org/x/text
golang.org/x/text v0.3.7
```
Closing this PR because the repo has already used v0.3.7, which fixed the CVE.

PR #268, opened 2022-08-12
Remove klusterlet agent namespace on the honsting cluster when delete klusterlet

zhujian7:
fix issue the klusterlet agent namespace will leak when delete the hosted mode klusterlet but there is no the external-managed-kubeconfig

### DIFF
--- a/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
+++ b/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
@@ -39,6 +39,9 @@ import (
 )
 
 const (
+
+	// klusterletHostedFinalizer is used to clean up resources on the managed/hosted cluster in Hosted mode
+	klusterletHostedFinalizer    = "operator.open-cluster-management.io/klusterlet-hosted-cleanup"
 	klusterletFinalizer          = "operator.open-cluster-management.io/klusterlet-cleanup"
 	imagePullSecret              = "open-cluster-management-image-pull-credentials"
 	klusterletApplied            = "Applied"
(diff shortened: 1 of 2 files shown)


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -274,50 +277,72 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
-		hasFinalizer := false
-		for i := range klusterlet.Finalizers {
-			if klusterlet.Finalizers[i] == klusterletFinalizer {
-				hasFinalizer = true
-				break
+	// Klusterlet is deleting, we remove its related resources on managed and management cluster
+	if !klusterlet.DeletionTimestamp.IsZero() {
+		var removeManagedClusterResources = true
+
+		if config.InstallMode == operatorapiv1.InstallModeHosted {
```
qiujian16: too many netsted if here.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -274,50 +277,72 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
 		config.InstallMode = operatorapiv1.InstallModeHosted
 	}
 
 	if config.InstallMode == operatorapiv1.InstallModeHosted {
+		var newStatus *operatorapiv1.KlusterletStatus
+		var updateErr error
 		managedClusterClients, err = n.buildManagedClusterClientsHostedMode(ctx, n.kubeClient, config.AgentNamespace, config.ExternalManagedKubeConfigSecret)
 		if err != nil {
-			_, _, _ = helpers.UpdateKlusterletStatus(ctx, n.klusterletClient, klusterletName, helpers.UpdateKlusterletConditionFn(metav1.Condition{
+			newStatus, _, updateErr = helpers.UpdateKlusterletStatus(ctx, n.klusterletClient, klusterletName, helpers.UpdateKlusterletConditionFn(metav1.Condition{
```
qiujian16: I wonder whether we just set the condition, and update at last? Also explain why we do not return err here.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -274,50 +277,72 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
-				break
+	// Klusterlet is deleting, we remove its related resources on managed and management cluster
+	if !klusterlet.DeletionTimestamp.IsZero() {
+		var removeManagedClusterResources = true
+
+		if config.InstallMode == operatorapiv1.InstallModeHosted {
+			if hasFinalezer(klusterlet, klusterletHostedFinalizer) {
+				if !meta.IsStatusConditionTrue(klusterlet.Status.Conditions, klusterletReadyToApply) {
+					// wait for the external managed kubeconfig to exist to clean up resources on the manged cluster
+					return nil
```
qiujian16: what will trigger this to reconcile, secret?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -274,50 +277,72 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
-		for i := range klusterlet.Finalizers {
-			if klusterlet.Finalizers[i] == klusterletFinalizer {
-				hasFinalizer = true
-				break
+	// Klusterlet is deleting, we remove its related resources on managed and management cluster
+	if !klusterlet.DeletionTimestamp.IsZero() {
+		var removeManagedClusterResources = true
+
+		if config.InstallMode == operatorapiv1.InstallModeHosted {
+			if hasFinalezer(klusterlet, klusterletHostedFinalizer) {
```
qiujian16: hasFinalizer

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -274,50 +277,72 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+
+		return n.removeKlusterletFinalizers(ctx, klusterlet)
 	}
 
-	// Klusterlet is deleting, we remove its related resources on managed cluster
-	if !klusterlet.DeletionTimestamp.IsZero() {
-		if err := n.cleanUp(ctx, controllerContext, managedClusterClients, config); err != nil {
-			return err
+	if config.InstallMode == operatorapiv1.InstallModeHosted {
+		if !meta.IsStatusConditionTrue(klusterlet.Status.Conditions, klusterletReadyToApply) {
```
qiujian16, line 331: why this cannot be put in 291?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: yes

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: Changed to using a switch statement, it is simpler now.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -465,6 +490,21 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+	for _, f := range klusterlet.Finalizers {
+		if f == finalizer {
+			return true
+		}
+	}
+	return false
+}
+
+func (n *klusterletController) addFinalizer(ctx context.Context, k *operatorapiv1.Klusterlet, finalizer string) error {
+	k.Finalizers = append(k.Finalizers, finalizer)
```
zhiweiyin318, line 514: need to check the finalizer has existed in the k.Finalizers ?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: it is not easy to determine where is the "last", it may return in different places.
And notes added to explain why we do not return here

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: because this klusterletReadyToApply condition is set after line 291, if we put it at line 291, this if statement will always be true.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: every caller has checked the finalizer is not present, otherwise, this func should not be invoked.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
qiujian16: if you put there, you do not need to check this condition. Only need to check the validity of the client, right?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -465,6 +485,37 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 	return nil
 }
 
+func skipCleanupManagedClusterResources(klusterlet *operatorapiv1.Klusterlet, mode operatorapiv1.InstallMode) bool {
```
qiujian16, line 488: add a comments here
1. we skip only in hosted mode and there is not hosted finalizer on klusterelt. The scenario happens when we cannot build managedclusterclient in hosted mode.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -274,50 +277,67 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
-		for i := range klusterlet.Finalizers {
-			if klusterlet.Finalizers[i] == klusterletFinalizer {
-				hasFinalizer = true
-				break
-			}
+	// Klusterlet is deleting, we remove its related resources on managed and management cluster
+	if !klusterlet.DeletionTimestamp.IsZero() {
+		skip := skipCleanupManagedClusterResources(klusterlet, config.InstallMode)
+
+		if !skip && !readyToCleanupManagedClusterResources(klusterlet, config.InstallMode) {
```
qiujian16, line 318: readyToCleanupManagedClusterResources seems not necessary, you just need to check the condition, right?

zhujian7:
Will trace by https://github.com/open-cluster-management-io/registration-operator/pull/269
/close

PR #269, opened 2022-08-17
Refactor to split two controllers to handle klusterlet deployment and cleanup

zhujian7:


qiujian16:
/approve

thanks, it looks much more neat.

PR #277, opened 2022-09-27
Fix: delete addon crd at first

qiujian16:
This is to ensure manifestwork of addon is deleted at first.


skeeey:
LGTM

PR #278, opened 2022-09-28
remove API Placement PlacementDecision ClusterSet ClusterSetBinding API version v1alpha1

haoqing0110:
Ref: https://github.com/open-cluster-management-io/community/issues/145

Signed-off-by: haoqing0110 <qhao@redhat.com>


deploy/cluster-manager/config/rbac/cluster_role.yaml
```diff
@@ -32,6 +32,10 @@ rules:
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
   verbs: ["create", "get", "list", "update", "watch", "patch", "delete"]
+# Allow the registration-operator to update crds status
```
haoqing0110, line 35: Need another PR in https://github.com/stolostron/backplane-operator/pull/297 to update the clusterrole.

haoqing0110 (commented):
Need another PR in https://github.com/stolostron/backplane-operator/pull/297 to update the clusterrole.

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -553,6 +566,40 @@ func applyManagementResources(
...
+		oldStoredVersions := crd.Status.StoredVersions
+		newStoredVersions := make([]string, 0, len(oldStoredVersions))
+		for _, stored := range oldStoredVersions {
+			if stored != version {
+				newStoredVersions = append(newStoredVersions, stored)
+			}
+		}
+		crd.Status.StoredVersions = newStoredVersions
+		// update the status sub-resource
+		crd, err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().UpdateStatus(ctx, crd, metav1.UpdateOptions{})
```
elgnay: No need to update the status of the CRD if there is no change at all.

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
```diff
@@ -113,18 +112,10 @@ func (c *crdMigrationController) sync(ctx context.Context, controllerContext fac
...
-		return removeStorageVersionMigrations(ctx, migrationClient)
-	}
-
-	// do not apply storage version migrations until other resources are applied
-	if applied := meta.IsStatusConditionTrue(clusterManager.Status.Conditions, clusterManagerApplied); !applied {
-		controllerContext.Queue().AddAfter(clusterManagerName, 5*time.Second)
-		return nil
-	}
-
-	return applyStorageVersionMigrations(ctx, migrationClient, controllerContext.Recorder())
```
elgnay: I think it's better for us to keep the logic of this controller, because there might be other CRDs that also need the storage version migration in future.

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
```diff
@@ -113,18 +112,10 @@ func (c *crdMigrationController) sync(ctx context.Context, controllerContext fac
...
+	// do not need to run applyStorageVersionMigrations since all the exisiting StorageVersionMigrations should be removed
+	// in this release.
+	// remove StorageVersionMigrations related resources on hub.
+	return removeStorageVersionMigrations(ctx, migrationClient)
```
elgnay: I remember we decided to keep the existing migration CRs. In another word, no new migration CRs will be created and leave the existing ones untouched.

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
haoqing0110: This function only updates CRDs defined in removedCRDStoredVersions

haoqing0110: Oh, I see, do you mean when running sync(), no need to update the status every time?

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
qiujian16: +1

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
qiujian16: we do not need to create new migration, and we do not need to remove the previous migration.

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
haoqing0110: Fixed

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
haoqing0110: Add the code back. Only remove the migration yamls and empty `migrationRequestFiles = []string{}`

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
haoqing0110: fixed

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -553,6 +567,43 @@ func applyManagementResources(
...
+			}
+		}
+
+		if !reflect.DeepEqual(oldStoredVersions, newStoredVersions) {
+			crd.Status.StoredVersions = newStoredVersions
+			// update the status sub-resource
+			crd, err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().UpdateStatus(ctx, crd, metav1.UpdateOptions{})
+			if err != nil {
+				klog.Warningf("faield to update CRD %v: %v", crd.Name, err)
+				continue
```
elgnay: It's not enough to output warning here. Instead, an error should be returned in this case, because the update of the crd (removing v1alpha1) will fail if v1alpha1 is still in the store version list.

haoqing0110: fixed

haoqing0110:
@qiujian16 @elgnay I've fixed all the comments, plz help review again when you are available, thank you.

PR #279, opened 2022-09-28
Add conversion webhook

ldpliu:
https://github.com/open-cluster-management-io/OCM/issues/67


pkg/helpers/helpers.go
```diff
@@ -335,6 +335,48 @@ func ApplyDeployment(
...
+		return fmt.Errorf("%q: %v", crdFile, err)
+	}
+	crd.(*apiextensionsv1.CustomResourceDefinition).Spec.Conversion = crdPatch.(*apiextensionsv1.CustomResourceDefinition).Spec.Conversion
+
+	if caBundle != "placeholder" {
+		//Inject ca bundle from ca-bundle-configmap
+		crd.(*apiextensionsv1.CustomResourceDefinition).Spec.Conversion.Webhook.ClientConfig.CABundle = []byte(caBundle)
+	}
+
+	_, _, err = resourceapply.ApplyCustomResourceDefinitionV1(
```
qiujian16: why don't we use patch directly?

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -493,6 +496,27 @@ func applyHubResources(
 		}
 	}
 
+	err = helpers.ApplyPatchedConversionCRD(
```
qiujian16: can we patch the manifest before apply?

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -493,6 +496,26 @@ func applyHubResources(
...
```
qiujian16: nit, let's add a comment here. I am still not sure whether it is a good approach. I think we could use a script to patch the manifest during make? The patch in controller make me feel it is still too complicated..

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
ldpliu: done

pkg/helpers/helpers.go
ldpliu: done

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
ldpliu: changed to use script to patch the crd

hack/patch/0000_00_clusters.open-cluster-management.io_managedclustersets.crd.yaml.yaml-patch
```diff
@@ -0,0 +1,16 @@
+#This is yaml-patch config file. It's used to patch conversion webhook config to clusterset crd
+- op: add
+  path: /spec/conversion
+  value:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          namespace: {{ .ClusterManagerNamespace }}
+          name: cluster-manager-registration-webhook
```
qiujian16, line 10: will it break cluster manager hosted mode? @xuezhaojun

qiujian16 (commented):
thanks, it looks neat. Only one comments that need @xuezhaojun to take a look.

PR #280, opened 2022-10-14
fix conversion webhook in hosted mode

ldpliu:
Signed-off-by: ldpliu <daliu@redhat.com>
https://github.com/open-cluster-management-io/OCM/issues/67

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,13 +116,13 @@ We mainly provide deployment in two scenarios:
    apiVersion: kind.x-k8s.io/v1alpha4
    nodes:
    - role: control-plane
-   extraPortMappings:
-   - containerPort: 30443
-      hostPort: 30443
-      protocol: TCP
-   - containerPort: 31443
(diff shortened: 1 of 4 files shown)


pkg/helpers/sa_syncer.go
```diff
@@ -155,7 +155,7 @@ func applyKubeconfigSecret(ctx context.Context, templateKubeconfig *rest.Config,
 		},
 		AuthInfos: map[string]*clientcmdapi.AuthInfo{
 			"user": {
-				TokenFile: "token",
+				Token: string(token),
```
qiujian16: why make this change?

xuezhaojun: Because the path is not right.

It could be fixed in this way:
```
TokenFile: "/var/run/secrets/hub/token"
```

qiujian16 (commented):
/approve

PR #287, opened 2022-10-26
correct managed cluster lease name

skeeey:
Signed-off-by: Wei Liu <liuweixa@redhat.com>


qiujian16:
hrm we might need to cherry-pick this.

PR #303, opened 2022-11-16
Make work agent feature gate configurable

zhujian7:

issue reference: https://github.com/open-cluster-management-io/work/issues/155

1. user can use the following config for the klusterlet to enable or disable the features for the work agent
```
spec:
  workConfiguration:
    featureGates:
    - feature: ExecutorValidatingCaches
      mode: Enable
```
2. refactor the condition updating in klusterlet controller to one place


zhujian7:
@qiujian16 @elgnay PTAL

pkg/helpers/helpers.go
```diff
@@ -173,12 +191,10 @@ func UpdateKlusterletConditionFn(conds ...metav1.Condition) UpdateKlusterletStat
 	}
 }
 
-// TODO: just for upgrading, we need to remove this in 0.11.0
-func ReplaceKlusterletConditionFn(removeConditionType string, conds ...metav1.Condition) UpdateKlusterletStatusFunc {
+func RemoveKlusterletConditionFn(condTypes ...string) UpdateKlusterletStatusFunc {
```
qiujian16, line 219: this should be removed in 0.11.0, right？

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -271,9 +227,11 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 
 	// If there are some invalid feature gates of registration, will output
 	// condition `ValidRegistrationFeatureGates` False in ClusterManager.
-	if clusterManager.Spec.RegistrationConfiguration != nil {
-		featureGates, condition, err := n.checkFeatureGate(ctx, clusterManagerName,
-			clusterManager.Spec.RegistrationConfiguration.FeatureGates, "registration")
+	if clusterManager.Spec.RegistrationConfiguration != nil &&
+		len(clusterManager.Spec.RegistrationConfiguration.FeatureGates) > 0 {
```
qiujian16: why we need to check empty? empty is a valid setting.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -242,52 +239,63 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+			helpers.UpdateKlusterletConditionFn(*conditions...))
+		if updateErr != nil {
+			klog.V(2).Infof("Update Klusterlet %q status with error: %v", klusterletName, updateErr)
+		}
+	}()
+
+	// If there are some invalid feature gates of registration, will output condition
+	// `ValidRegistrationFeatureGates` False in Klusterlet.
+	if klusterlet.Spec.RegistrationConfiguration != nil &&
+		len(klusterlet.Spec.RegistrationConfiguration.FeatureGates) > 0 {
```
qiujian16: we should consider empty case

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -242,52 +239,63 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
-			_, _, _ = helpers.UpdateKlusterletStatus(ctx, n.klusterletClient, klusterletName, helpers.ReplaceKlusterletConditionFn(
-				spokeRegistrationFeatureGatesInvalid,
-				metav1.Condition{
-					Type: spokeRegistrationFeatureGatesValid, Status: metav1.ConditionFalse, Reason: "InvalidFeatureGatesExisting",
-					Message: invalidFGErr.Error(),
-				}))
-			return invalidFGErr
+	// Do not use "conditions := &klusterlet.Status.Conditions", otherwise will override conditions
+	// set by other controllers
+	conditions := &[]metav1.Condition{}
```
qiujian16: why?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: so, if the feature gates are empty, do you expect to set a valid condition=true condition?

pkg/helpers/helpers.go
zhujian7: yes

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: the `conditions` we set here likely mean "additional" conditions.
For example, if we use `conditions := &klusterlet.Status.Conditions`, at the beginning there is a condition `WorkDesiredDegraded=true` in the `klusterlet.Status.Conditions`, and another controller updates a condition `WorkDesiredDegraded=falst` **at the same time**, and at the end of this `sync` we update the status using `UpdateKlusterletStatus` which will update WorkDesiredDegraded condition back to `true`.

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
zhujian7: when it is empty, I  think we do not need to set the condition for `ValidRegistrationFeatureGates` or `ValidWorkFeatureGates`

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
qiujian16: we should not have two controllers update the same condition type.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
qiujian16: yes

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
qiujian16: I think condition should always be set. And if user set feature is empty. We need to change the deployment config. It is different from nil.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: yes, that is why we do not use `conditions := &klusterlet.Status.Conditions` here,  we use `conditions := &[]metav1.Condition{}` to set conditions that are only generated by the current controller.

pkg/helpers/helpers.go
```diff
@@ -887,11 +916,126 @@ func FeatureGatesArgs(featureGates []operatorapiv1.FeatureGate, component string
...
+				Message: fmt.Sprintf("%s feature gates are all valid", componentName),
+			}, nil
+		}
+
+		// for the invalid feature gates, we show them in the condition, here only return the valid
+		// feature gates, so the caller can use the valid ones and degrade the invalid ones to the
+		// default values
+		return featureGateArgs, metav1.Condition{
+			Type: conditionType, Status: metav1.ConditionFalse, Reason: "InvalidFeatureGatesExisting",
+			Message: fmt.Sprintf("there are some invalid feature gates of %s: %v, will process them with default values",
```
zhujian7: if the feature gate is invalid, will use the default value to deploy the registration/work-agent, add an invalid condition.
@qiujian16 Do you think this is reasonable?

zhujian7:
@qiujian16 @zhiweiyin318  PTAL

pkg/helpers/helpers.go
qiujian16: what if one feature is correct, and another is invalid?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -268,6 +238,29 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		return nil
 	}
 
+	// If there are some invalid feature gates of registration, will output condition
+	// `ValidRegistrationFeatureGates` False in Klusterlet.
+	featureGates, condition, err := helpers.CheckRegistrationFeatureGates(helpers.OperatorTypeKlusterlet,
```
qiujian16: why not return work and registration feature at the same time?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -268,6 +238,29 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
...
+
+	// If there are some invalid feature gates of work, will output condition
+	// `ValidWorkFeatureGates` False in Klusterlet.
+	// TODO: When splitting permissions in the future, if the ExecutorValidatingCaches function is enabled,
+	//       additional permissions for get, list, and watch RBAC resources required by this function need
+	//       to be applied
+	featureGates, condition, err = helpers.CheckWorkFeatureGates(helpers.OperatorTypeKlusterlet,
+		klusterlet.Spec.WorkConfiguration)
+	if err != nil {
+		return err
```
qiujian16: should we update condition here?

pkg/helpers/helpers.go
```diff
@@ -887,11 +916,126 @@ func FeatureGatesArgs(featureGates []operatorapiv1.FeatureGate, component string
...
+		return nil, metav1.Condition{}, fmt.Errorf("not supported operatorType: %s", operatorType)
+	}
+
+	if configuration == nil {
+		return checkNilFeatureGates(operatorType, componentNameWork)
+	}
+	return checkFeatureGates(operatorType, componentNameWork, configuration.FeatureGates)
+}
+
+func checkNilFeatureGates(operatorType string, componentName string) ([]string, metav1.Condition, error) {
```
qiujian16: not sure this condition reason is helpful. I think nil is a valid, and we just return common valid reason

pkg/helpers/helpers.go
```diff
@@ -887,11 +916,126 @@ func FeatureGatesArgs(featureGates []operatorapiv1.FeatureGate, component string
...
+	}
+
+	return componentKey
+}
+
+func getConditionType(componentName string) string {
+	var conditionType string
+	switch componentName {
+	case componentNameRegistration:
+		conditionType = RegistrationFeatureGatesValid
```
qiujian16: hrm, I am thinking why we need to conditions. Shouldn't we have one condition and different reasons?

pkg/helpers/helpers.go
zhujian7: the condition will be invalid, and print the invalid feature gate. And will use the user-provided value for the correct one and the default value for the invalid one to deploy registration/work agent.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: With that, there are 5 return values, which is a bit too many for me.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: it is not necessary, an error will be returned only when the wrong operatorType is passed in, and this case does not exist. Maybe I need to remove the `err` return value here.

pkg/helpers/helpers.go
zhujian7: You mean merge the RegistrationFeatureGatesValid and WorkFeatureGatesValid into one condition？

pkg/helpers/helpers.go
zhujian7: OK, so do you think the reason FeatureGatesEmpty is necessary？

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
zhujian7: if we remove the `error` return value, and merge these two, it would be
```
CheckFeatureGates(operatorType string, registrationConfiguration *operatorapiv1.RegistrationConfiguration,
	workConfiguration *operatorapiv1.WorkConfiguration) (registrationFeatureGates []string, 
	workFeatureGates []string, registrationCondition metav1.Condition, workCondition metav1.Condition)
```
I feel it's a little bit more complicated

pkg/helpers/helpers.go
qiujian16: no it is not

pkg/helpers/helpers.go
qiujian16: yes, I do not see why we need two conditions.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
qiujian16: I think we should have only one condition.

zhujian7:
@qiujian16 all comments have been addressed. PTAL

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -192,28 +148,17 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
...
-			return err
-		}
-		config.WorkFeatureGates = featureGates
-		meta.SetStatusCondition(&clusterManager.Status.Conditions, condition)
-	}
+	var conds []metav1.Condition
+
+	// If there are some invalid feature gates of registration or work, will output
+	// condition `ValidFeatureGates` False in ClusterManager.
+	registrationFeatureGates, workFeatureGates, featureGateCondition := helpers.CheckFeatureGates(
```
qiujian16: nit, I would do 

config.RegistrationFeatureGates, config.WorkFeatureGates, featureGateCondition = ...

pkg/helpers/helpers.go
```diff
@@ -140,6 +173,14 @@ func UpdateKlusterletStatus(
 		oldStatus := &klusterlet.Status
 
 		newStatus := oldStatus.DeepCopy()
+		// TODO: just for upgrading, we change the condition type from "InvalidRegistrationFeatureGates" to
+		// "ValidRegistrationFeatureGates", need to remove this in 0.11.0
+		if err := RemoveKlusterletConditionFn("InvalidRegistrationFeatureGates",
+			"ValidRegistrationFeatureGates", "ValidWorkFeatureGates")(newStatus); err != nil {
```
qiujian16: do we have ValidWorkFeatureGates, ValidRegistrationFeatureGates already? I think we introduce it recently, and we have not released 0.10.0

zhujian7: [clusterManager](https://github.com/open-cluster-management-io/registration-operator/blob/v0.9.0/pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go#L141-L142) has ValidWorkFeatureGates, ValidRegistrationFeatureGates in 0.9.0
and [klusterlet](https://github.com/open-cluster-management-io/registration-operator/blob/v0.9.0/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go#L249) has ValidRegistrationFeatureGates and a wrong InvalidRegistrationFeatureGates, but does not have ValidWorkFeatureGates in 0.9.0. so I will delete the removing of ValidWorkFeatureGates for klusterlet, but keep others.

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
zhujian7: Fixed

PR #306, opened 2022-11-22
Refactor klusterlet

qiujian16:

### DIFF
--- a/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
+++ b/pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
@@ -16,23 +16,18 @@ import (
 	appsinformer "k8s.io/client-go/informers/apps/v1"
 	coreinformer "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
-	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
-	"github.com/openshift/library-go/pkg/assets"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
-	operatorhelpers "github.com/openshift/library-go/pkg/operator/v1helpers"
-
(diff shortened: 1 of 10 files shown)


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_pullsecret_reconcile.go
```diff
@@ -0,0 +1,119 @@
...
+	// Sync pull secret to the klusterlet addon namespace
+	// The reason we keep syncing secret instead of adding a label to trigger addonsecretcontroller to sync is:
+	// addonsecretcontroller only watch namespaces in the same cluster klusterlet is running on.
+	// And if addons are deployed in default mode on the managed cluster, but klusterlet is deployed in hosted on management cluster, then we still need to sync the secret here in klusterlet-controller using `managedClusterClients.kubeClient`.
+	err = r.syncPullSecret(ctx, r.kubeClient, r.managedClusterClients.kubeClient, klusterlet, addonNamespace, r.recorder)
+	if err != nil {
+		return klusterlet, reconcileStop, err
+	}
+
+	if config.InstallMode == operatorapiv1.InstallModeHosted {
```
zhujian7: Is it better to put this into the `managedReconcile`?

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -117,6 +67,17 @@ type klusterletController struct {
...
 }
 
+type klusterletReconcile interface {
+	reconcile(ctx context.Context, cm *operatorapiv1.Klusterlet, config klusterletConfig) (*operatorapiv1.Klusterlet, reconcileState, error)
+}
+
+type reconcileState int64
+
+const (
+	reconcileStop reconcileState = iota
```
zhujian7, line 77: Why do we need this? I observed that every use of reconcileStop is accompanied by an error

qiujian16: not necessarily, it is possible that we continue to run another reconcile with error. e.g. apply anyway but still return error so it check again later.

pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_pullsecret_reconcile.go
qiujian16: yes, this reconcile is confusing, updated.

PR #309, opened 2022-12-08
Add ClusterManager CR uninstall case in e2e.

xuezhaojun:

Belongs to https://github.com/stolostron/backlog/issues/25657.

### DIFF
--- a/test/e2e/clustermanager_test.go
+++ b/test/e2e/clustermanager_test.go
@@ -1,0 +1,64 @@
+package e2e
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"golang.org/x/net/context"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	operatorapiv1 "open-cluster-management.io/api/operator/v1"
+)
+
+// clustermanager is used to store and receate the ClusterManager CR.
+var clustermanager *operatorapiv1.ClusterManager
+
+// This ordered describe is only used to test ClusterManager CR, it can not run in parallel with other tests.
+// Please run `make test-e2e-in-parallel` in order to skip this case if needed.
+// Also please add other testcases of cluster manager to a separate describe block.
+var _ = Describe("Delete and recreate ClusterManager CR", Label("run-alone"), Ordered, func() {
+	When("ClusterManager CR is deleted", func() {
+		It("Should remove CRDs and namespace", func() {
+			var err error
+
+			// Cache the current ClusterManager CR
+			clustermanager, err = t.GetClusterManager(context.TODO())
+			Expect(err).To(BeNil())
+
+			// Delete ClusterManager CR
+			err = t.DeleteClusterManager(context.TODO())
+			Expect(err).To(BeNil())
+
+			// Check namespace are removed
+			Eventually(func() error {
+				_, err = t.KubeClient.CoreV1().Namespaces().Get(context.TODO(), t.clusterManagerNamespace, metav1.GetOptions{})
+				if err != nil {
+					if errors.IsNotFound(err) {
+						return nil
+					}
+					return err
+				}
+				return fmt.Errorf("namespace %s still exists", t.clusterManagerNamespace)
+			}, t.EventuallyTimeout, t.EventuallyInterval).Should(Succeed())
+		})
+	})
+
+	When("ClusterManager CR is recreated", func() {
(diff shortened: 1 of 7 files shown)


test/integration/clustermanager_hosted_test.go
```diff
@@ -81,7 +81,7 @@ var _ = ginkgo.Describe("ClusterManager Hosted Mode", func() {
 		}
 	})
 
-	ginkgo.Context("Deploy and clean hub component", func() {
```
xuezhaojun, line 84: Remove "and clean" in the intergration-test context to make it more clear.

test/e2e/common.go
```diff
@@ -484,45 +503,41 @@ func (t *Tester) CheckHubReady() error {
...
 		Get(context.TODO(), t.hubRegistrationDeployment, metav1.GetOptions{}); err != nil {
+		return fmt.Errorf("failed to get deployment %s in %s: %v", t.hubRegistrationDeployment, t.clusterManagerNamespace, err)
+	}
+	// registration-webhook
+	registrationWebhookDeployment, err := t.KubeClient.AppsV1().Deployments(t.clusterManagerNamespace).
+		Get(context.TODO(), t.hubRegistrationWebhookDeployment, metav1.GetOptions{})
+	if err != nil {
 		return err
 	}
-	gomega.Eventually(func() error {
```
xuezhaojun: Remove `Eventually` inside of the `CheckHubReady` function, since the function is already used with `Eventually` outside.

test/e2e/clustermanager_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"golang.org/x/net/context"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	operatorapiv1 "open-cluster-management.io/api/operator/v1"
+)
+
+// clustermanager is used to store and receate the ClusterManager CR.
+var clustermanager *operatorapiv1.ClusterManager
+
+var _ = Describe("Delete and recreate ClusterManager CR", Ordered, func() {
```
xuezhaojun: Here we use `Ordered` to make `recreate` happen after `delete`.

xuezhaojun:
/hold
The test case of removing clustermanager would affect other test cases. This one needs further discussion.

test/e2e/clustermanager_test.go
qiujian16: we need to add a comment indicating when it should be run

test/e2e/clustermanager_test.go
```diff
@@ -0,0 +1,61 @@
...
+	When("ClusterManager CR is deleted", func() {
+		It("Should remove CRDs and namespace", func() {
+			var err error
+
```
qiujian16, line 25: we need to add a conditional skip so we can choose not to run sometime.

xuezhaojun:
@qiujian16 Please take another look. Thanks!
Created a new directory `e2eordered` for all cases that need to run in order. We will run `e2eordered` before `e2e` to make cases like `uninstall clustermanager CR` won't be affected by other cases.

qiujian16:
it is an uncommon pattern...Instead, can we just create/clean clustermanager per test cases? or just make this conditional skipped.

xuezhaojun:
> it is an uncommon pattern...Instead, can we just create/clean clustermanager per test cases? or just make this conditional skipped.

Add create/clean per test case means we need to add `BeforeEach` function to every `*_test.go` now, and every time we want to add a new case in the future. It feels a little too much for this task.
And I'm not sure I catch the idea of conditional skipping correctly, Does that mean there are chances that we will miss this case in a PR?

---

And I found a way to solve the problem in our initial version solution which is we worry about some other cases may change the system's state and leads to the clustermanager CR deletion blocked.

I found ginkgo has a feature [serial specs](https://onsi.github.io/ginkgo/#serial-specs). For cases that lead to a bad state (In our case, can't recreate clustermanager CR), we can add `Serial` on them, and ginkgo can make sure those cases run **after all others cases are done**.

Based on this feature, do we consider bringing the initial solution back to the table? Like, if some cases blocks others, we make them run at the end by adding `Serial`.

test/e2e/clustermanager_test.go
xuezhaojun: Comment added.

test/e2e/clustermanager_test.go
xuezhaojun: We are going to use ginkgo's `Label` to filter cases in run.

xuezhaojun:
/close
Since we would merge different repos into one and testcases must be merged as well.

PR #311, opened 2023-01-12
Allow customizing the klusterlet name when deploying in hosted mode

mprahl:
This will allow users of this Makefile to deploy multiple managed clusters in hosted mode.


qiujian16:
https://github.com/open-cluster-management-io/clusteradm/pull/289 this might also be related.

PR #312, opened 2023-01-18
cleanup return if appliedmainfestWork is not found

zhiweiyin318:


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_managed_reconcile.go
```diff
@@ -168,6 +168,9 @@ func (r *managedReconcile) clean(ctx context.Context, klusterlet *operatorapiv1.
 func (r *managedReconcile) cleanUpAppliedManifestWorks(ctx context.Context, klusterlet *operatorapiv1.Klusterlet, config klusterletConfig) error {
 	appliedManifestWorks, err := r.managedClusterClients.appliedManifestWorkClient.List(ctx, metav1.ListOptions{})
 	if err != nil {
+		if errors.IsNotFound(err) {
```
qiujian16:
```go
if errors.IsNotFound(err) {
	return nil
}
if err != nil {
	return fmt.Errorf("unable to list AppliedManifestWorks: %w", err)
}
```

zhiweiyin318: done

PR #316, opened 2023-02-09
regenerate and load hosted secrets when ca changes

zhujian7:
- check if the host and ca chengs in the `external-managed-kubeconfig` secret, if it does, regenerate the `external-managed-kubeconfig-registration` and `external-managed-kubeconfig-work`
- use the `--terminate-on-files` flag to watch the hub/spoke kubeconfig file, if there is any change, restart the pod. there is one side effect for the registration-agent, before this change: the regisatrion can reload the hub kubeconfig when it is generated first time(importing cluster, bootstrap secret -> hub kubeconfig secret) without restarting pod. but with the change, the registration-agent will restart. if we want to prevent the side effect, we can use https://github.com/open-cluster-management-io/registration/pull/295, but it is complicated.


qiujian16 (commented):
/approve
/lgtm

PR #319, opened 2023-02-15
fallback to get hub host from hub kubeconfig secret when detaching cluster

zhujian7:


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_managed_reconcile.go
```diff
@@ -178,17 +178,16 @@ func (r *managedReconcile) cleanUpAppliedManifestWorks(ctx context.Context, klus
 		return nil
 	}
 
-	bootstrapKubeConfigSecret, err := r.kubeClient.CoreV1().Secrets(config.AgentNamespace).Get(ctx, config.BootStrapKubeConfigSecret, metav1.GetOptions{})
+	hubHost, err := r.getHostFromKubeconfigSecret(ctx, config.AgentNamespace, config.BootStrapKubeConfigSecret)
```
qiujian16, line 181: this is tricky...what happens when all secret are missing? Actually should we use agentID only to handle applied manifestwork regardless of hubhash, what is the side effect?

zhujian7: I think we just introduced the `appliedmanifestwork.spec.agentID` last release, I'm not sure all appliedmanifestwork, upgrading case, will have the value

qiujian16: work agent will update agent id for each appliedmanifestwork

zhujian7:
https://github.com/open-cluster-management-io/registration-operator/pull/321 will fix this.
/close

PR #325, opened 2023-02-23
addon manager feature gates

haoqing0110:


pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -211,6 +217,18 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 		return removeClusterManagerFinalizer(ctx, n.clusterManagerClient, clusterManager)
 	}
 
+	// If AddOnManager is not enabled, remove related resources
+	if operatorapiv1.ComponentModeType(config.AddOnManagerComponentMode) != operatorapiv1.ComponentModeTypeEnable {
```
qiujian16: can we do it in reconciler? I do not think we should clean it in the level.

haoqing0110: yes, moved to reconcile().

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -61,7 +62,7 @@ type clusterManagerController struct {
 
 type clusterManagerReconcile interface {
 	reconcile(ctx context.Context, cm *operatorapiv1.ClusterManager, config manifests.HubConfig) (*operatorapiv1.ClusterManager, reconcileState, error)
-	clean(ctx context.Context, cm *operatorapiv1.ClusterManager, config manifests.HubConfig) (*operatorapiv1.ClusterManager, reconcileState, error)
+	clean(ctx context.Context, cm *operatorapiv1.ClusterManager, config manifests.HubConfig, resource ...string) (*operatorapiv1.ClusterManager, reconcileState, error)
```
qiujian16: I do not think we should the change the signature

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
```diff
@@ -149,6 +166,7 @@ func (c *runtimeReconcile) reconcile(ctx context.Context, cm *operatorapiv1.Clus
 func (c *runtimeReconcile) clean(ctx context.Context, cm *operatorapiv1.ClusterManager, config manifests.HubConfig) (*operatorapiv1.ClusterManager, reconcileState, error) {
 	// Remove All Static files
 	managementResources := []string{namespaceResource} // because namespace is removed, we don't need to remove deployments explicitly
+
```
qiujian16: can we call cleanResource here?

pkg/helpers/helpers.go
```diff
@@ -728,6 +744,18 @@ func FindRelatedResourcesStatus(
 	return nil
 }
 
+func RemoveRelatedResourcesStatus(
+	relatedResourcesStatuses *[]operatorapiv1.RelatedResourceMeta,
+	relatedResource operatorapiv1.RelatedResourceMeta) {
+	var result []operatorapiv1.RelatedResourceMeta
+	for _, v := range *relatedResourcesStatuses {
+		if !reflect.DeepEqual(v, relatedResource) {
```
qiujian16: relatedResource is a plain struct that you can directly use == I think

pkg/helpers/helpers.go
```diff
@@ -717,6 +719,20 @@ func SetRelatedResourcesStatuses(
 	}
 }
 
+func RemoveRelatedResourcesStatuses(
```
qiujian16, line 711: this needs tests. And are you sure we need this? My understanding is everytime we will update the whole relatedResource list.

haoqing0110: Hmm, I think the existing func `SetRelatedResourcesStatuses` and this new func `RemoveRelatedResourcesStatuses` are to add/remove obj to/from the relatedResource when applying/deleting resource. They are running before `UpdateClusterManagerRelatedResourcesFn` (which is to update the whole list you mentioned here). 

Will add more testing code.

haoqing0110: Add some UT and add relatedResources check in integration testing.

pkg/helpers/helpers.go
haoqing0110: fixed

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
haoqing0110: fixed

PR #330, opened 2023-03-01
remove old webhook

ldpliu:


qiujian16 (commented):
for upgrade I think we already remove apiservices in the last release?

ldpliu:
> for upgrade I think we already remove apiservices in the last release?

Yes

qiujian16 (commented):
/approve
/lgtm

PR #331, opened 2023-03-02
Add manifestWorkReplicaSets controller

serngawy:


manifests/cluster-manager/hub/cluster-manager-registration-clusterrole.yaml
```diff
@@ -42,6 +42,10 @@ rules:
 - apiGroups: ["work.open-cluster-management.io"]
   resources: ["manifestworks"]
   verbs: ["get", "list", "watch"]
+# Allow hub to monitor manifestworkreplicasets
+- apiGroups: ["work.open-cluster-management.io"]
```
qiujian16, line 46: you also need manifestworkreplicasets/status, and manifestworkreplicasets/finalize permission I think

qiujian16 (commented):
I think we need a flag on clustermanager api to enable it, and we should disable it by default in the alpha stage. like this one https://github.com/open-cluster-management-io/api/pull/211

manifests/cluster-manager/hub/cluster-manager-registration-clusterrole.yaml
serngawy: done

serngawy:
> I think we need a flag on clustermanager api to enable it, and we should disable it by default in the alpha stage. like this one [open-cluster-management-io/api#211](https://github.com/open-cluster-management-io/api/pull/211)

okay, but the manifestWorkReolicaSet it is already part of the work repo I believe its belong to workConfiguration . is it will be as [FeatureGate](https://github.com/open-cluster-management-io/api/blob/main/operator/v1/types.go#L88) ?  if yes, what name it should has ?

qiujian16:
maybe a featureGate EnableReplicaSetManager

serngawy:
> maybe a featureGate EnableReplicaSetManager

fix it, it has nothing to do with the featureGate , the integration test failed due to number of CRDs created.

qiujian16 (commented):
you would also need to start the manager on the hub side. we currently only start webhook.

serngawy:
> you would also need to start the manager on the hub side. we currently only start webhook.

Would you advice Jian I'm kind of lost between the work repo and this one.
 I added the manager-deployment and the manifestWorkReplicaSet-webhookvalidation. Do we need to add service and endpoint as well ?

qiujian16:
no, we do not need service

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+        args:
+          - "/work"
+          - "manager"
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
+          privileged: false
+          runAsNonRoot: true
```
qiujian16, line 58: also need livenessprobe/readinessprobe and resource request. We also need image part as a template variable.

manifests/cluster-manager/hub/cluster-manager-manifestworkreplicaset-webhook-validatingconfiguration-v1alpha1.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: manifestworkreplicasetv1alpha1validators.admission.work.open-cluster-management.io
+webhooks:
+- name: manifestworkreplicasetv1alpha1validators.admission.work.open-cluster-management.io
+  failurePolicy: Fail
+  clientConfig:
+    service:
+      namespace: {{ .ClusterManagerNamespace }}
+      name: cluster-manager-registration-webhook
```
qiujian16: I think the service name should be cluster-manager-work-webhook

manifests/cluster-manager/hub/cluster-manager-manifestworkreplicaset-webhook-validatingconfiguration.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: admissionregistration.k8s.io/v1
```
qiujian16, line 1: why we need two same configurations?

serngawy: ops , forget to delete it

pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
```diff
@@ -29,6 +29,7 @@ var (
 		"cluster-manager/management/cluster-manager-registration-webhook-deployment.yaml",
 		"cluster-manager/management/cluster-manager-work-webhook-deployment.yaml",
 		"cluster-manager/management/cluster-manager-placement-deployment.yaml",
+		"cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml",
```
zhiweiyin318, line 32: should we check the featureGate EnableReplicaSetManager to determine if install the manifestworkreplicaset manifests like  https://github.com/open-cluster-management-io/registration-operator/blob/547fc1482e257e7eb8abfcd13c456513837d5574/pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go#L106 ?

qiujian16:
@zhiweiyin318 let's merge it at first, We could handle the comment as the followup

/approve
/lgtm

PR #332, opened 2023-03-07
refactor migration and storedversion update

ldpliu:


pkg/operators/clustermanager/controllers/crdstatuccontroller/crd_status_controller.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+var (
+	// removed CRD StoredVersions
+	removedCRDStoredVersions = map[string]string{
+		"managedclustersets.cluster.open-cluster-management.io":        "v1beta1",
+		"managedclustersetbindings.cluster.open-cluster-management.io": "v1beta1",
+	}
+	// latest required CRD StoredVersions
+	requiredCRDStoredVersion = map[string]string{
```
qiujian16: we only need the required version I think

pkg/operators/clustermanager/controllers/crdstatuccontroller/crd_status_controller.go
```diff
@@ -0,0 +1,180 @@
...
+		"managedclustersets.cluster.open-cluster-management.io":        "v1beta2",
+		"managedclustersetbindings.cluster.open-cluster-management.io": "v1beta2",
+	}
+)
+
+type crdStatusController struct {
```
qiujian16, line 38: why it is not a reconciler in clustermanager controller? Do you expect this to be a more generic controller that can be reused in the future?

ldpliu: 1. I think clustermanager controller is used to apply basic resources but crdStatusController works only after cluster manager applied succeed. 
2. I do not want crdStatusController  affect `Applied condition` in clustermanager, and if it fails, it will just need to try reconcile its own logic, do not need to apply other resousces in clustermanager controller logic.

pkg/operators/clustermanager/controllers/crdstatuccontroller/crd_status_controller.go
ldpliu: changed

pkg/operators/clustermanager/controllers/crdstatuccontroller/crd_status_controller.go
```diff
@@ -0,0 +1,173 @@
...
+// updateStoredVersion update(remove) deleted api version from CRD status.StoredVersions
+func updateStoredVersion(ctx context.Context, hubAPIExtensionClient apiextensionsclient.Interface) error {
+	for name, desiredVersions := range desiredCRDStoredVersions {
+		// retrieve CRD
+		crd, err := hubAPIExtensionClient.ApiextensionsV1().CustomResourceDefinitions().Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		isInclude := includeStrings(crd.Status.StoredVersions, desiredVersions)
```
qiujian16: I think we just need deepequal?

ldpliu: Changed.
As the `StoredVersions` is a `[]string`, and deepequal can not  used here directly. 
So I use a `sets.string` to convert the `[]string`,  then use deepequal to check it.

qiujian16: Curious: why we cannot use reflect.DeepEqual against []string?

ldpliu: > Curious: why we cannot use reflect.DeepEqual against []string?

For example:
```
s1[0]="v1beta1", s1[1]="v1beta2"
s2[0]="v1beta2", s2[1]="v1beta1"
```

In our Scenario, s1 should be same as s2. but reflect.DeepEqual(s1,s2)  will return false.

qiujian16: How does string array be disordered? I think kube crd controller will update storedversion with an ordered string array.

ldpliu: You are right. 
Changed to use reflect.DeepEqual to compare the array directly.

qiujian16 (commented):
/approve

qiujian16:
maybe @elgnay or @haoqing0110 need to take a look also

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
```diff
@@ -174,70 +208,129 @@ func removeStorageVersionMigrations(
...
-		return nil, false, fmt.Errorf("invalid StorageVersionMigration in file %q: %v", file, requiredObj)
+		return nil, fmt.Errorf("invalid StorageVersionMigration in file %q: %v", file, svmObj)
 	}
 
+	return svm, nil
+}
+
+func applyStorageVersionMigration(
+	client migrationv1alpha1client.StorageVersionMigrationsGetter,
+	required *migrationv1alpha1.StorageVersionMigration,
```
haoqing0110, line 331: Seems the required comes from `parseStorageVersionMigrationFile`. Need to check required is nil here?

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
```diff
@@ -103,27 +113,33 @@ func (c *crdMigrationController) sync(ctx context.Context, controllerContext fac
...
+		migrationCond := metav1.Condition{
+			Type:    MigrationSucceeded,
+			Status:  metav1.ConditionFalse,
+			Reason:  "StorageVersionMigrationFailed",
+			Message: fmt.Sprintf("Do not support StorageVersionMigration"),
+		}
+		_, _, err = helpers.UpdateClusterManagerStatus(ctx, c.clusterManagerClient, clusterManagerName,
+			helpers.UpdateClusterManagerConditionFn(migrationCond),
+		)
+		return err
```
haoqing0110, line 141: Does this mean if not supported, the condition will always be false? and CRD stored version will never be removed?

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
```diff
@@ -174,70 +208,129 @@ func removeStorageVersionMigrations(
...
+			func(name string) ([]byte, error) {
+				template, err := manifests.ClusterManagerManifestFiles.ReadFile(name)
+				if err != nil {
+					return nil, err
+				}
+				return assets.MustCreateAssetFromTemplate(name, template, struct{}{}).Data, nil
+			},
+			file)
+		if err != nil {
+			return metav1.Condition{}, err
```
haoqing0110, line 256: Same as above question, does this mean if no required migration file, there will be no condition? and the crd_status_controller will keep requeueing every 5 seconds?

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
ldpliu: Yes, user need to manually delete it or add a `MigrationSucceeded` condition in clustermanger status condition.

pkg/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
haoqing0110: Also, if not supported, will the condition always be false and crd_status_controller will keep requeueing every 5 seconds?

PR #334, opened 2023-03-21
Add e2e test on lower kubernetes version.

xuezhaojun:


.github/workflows/go-presubmit.yml
```diff
@@ -117,11 +149,11 @@ jobs:
       - name: setup kind
         uses: engineerd/setup-kind@v0.5.0
         with:
-          version: v0.17.0
-          name: cluster1
+          version: v0.6.0 # The default node image is a Kubernetes v1.16.3, see https://github.com/kubernetes-sigs/kind/releases/tag/v0.6.0
```
qiujian16, line 152: We do not need to support hub with v1.16.3

xuezhaojun: Got it.

xuezhaojun:
/close 
As we decided to add regression tests.

PR #335, opened 2023-03-23
fix Implicit memory aliasing in for loop

ldpliu:


pkg/operators/klusterlet/controllers/klusterletcontroller/klusterlet_managed_reconcile.go
```diff
@@ -178,18 +178,18 @@ func (r *managedReconcile) cleanUpAppliedManifestWorks(ctx context.Context, klus
 	}
 
 	var errs []error
-	for _, appliedManifestWork := range appliedManifestWorks.Items {
+	for index, appliedManifestWork := range appliedManifestWorks.Items {
```
qiujian16: should this be `for index := range appliedManifestWorks.Items`?

ldpliu: > should this be `for index := range appliedManifestWorks.Items`?

done

PR #339, opened 2023-03-30
add e2e for deleting klusterlet when the managed cluster was destroyed

zhujian7:


zhujian7:
cc @qiujian16

qiujian16 (commented):
/approve
/lgtm

PR #340, opened 2023-03-30
Fix ManifestWorkReplicaset featureGate

serngawy:


qiujian16:
I am fixing another featuregate here https://github.com/open-cluster-management-io/registration-operator/pull/338 I think you could later use the similar way

serngawy:
> I am fixing another featuregate here #338 I think you could later use the similar way

okay, sounds like same path

deploy/cluster-manager/config/samples/operator_open-cluster-management_clustermanagers.cr.yaml
```diff
@@ -13,3 +13,8 @@ spec:
     featureGates:
     - feature: DefaultClusterSet
       mode: Enable
+  # Enable ManifestWorkReplicaSet for e2e
+  workConfiguration:
```
qiujian16: we should not by default enable it in the example

qiujian16 (commented):
/approve

looks good in general, just a minor comment

deploy/cluster-manager/config/samples/operator_open-cluster-management_clustermanagers.cr.yaml
serngawy: okay, removed.

PR #344, opened 2023-04-11
remove escalate bind rbac

zhiweiyin318:


deploy/cluster-manager/config/rbac/cluster_role.yaml
```diff
@@ -55,3 +58,65 @@ rules:
...
+  verbs: ["renew"]
+- apiGroups: ["work.open-cluster-management.io"]
+  resources: ["manifestworkreplicasets"]
+  verbs: ["get", "list", "watch", "update", "delete", "deletecollection", "patch"]
+- apiGroups: ["work.open-cluster-management.io"]
+  resources: ["manifestworkreplicasets/finalizers"]
+  verbs: ["update"]
+- apiGroups: ["work.open-cluster-management.io"]
+  resources: ["manifestworks"]
+  verbs: ["get", "list", "watch", "create", "update", "delete", "deletecollection", "patch", "execute-as"]
```
qiujian16, line 126: also need create verb for manifestworkreplicaset controller

deploy/cluster-manager/config/rbac/cluster_role.yaml
```diff
@@ -55,3 +58,65 @@ rules:
 - apiGroups: ["migration.k8s.io"]
   resources: ["storageversionmigrations"]
   verbs: ["create", "get", "list", "update", "watch", "patch", "delete"]
+
+- apiGroups: ["addon.open-cluster-management.io"]
+  resources: ["managedclusteraddons"]
+  verbs: ["get", "list", "watch"]
```
qiujian16: create/update/patch is also need for addon-manager controller

deploy/cluster-manager/config/rbac/cluster_role.yaml
zhiweiyin318: added

deploy/cluster-manager/config/rbac/cluster_role.yaml
zhiweiyin318: added

PR #347, opened 2023-04-19
Add gosec in verify process.

xuezhaojun:


pkg/helpers/queuekey.go
```diff
@@ -37,7 +37,7 @@ const (
 
 	RegistrationWebhookSecret  = "registration-webhook-serving-cert"
 	RegistrationWebhookService = "cluster-manager-registration-webhook"
-	WorkWebhookSecret          = "work-webhook-serving-cert"
+	WorkWebhookSecret          = "work-webhook-serving-cert" // #nosec G101
```
qiujian16, line 40: what does this nosec G101 mean?

xuezhaojun: G101 means gosec think it may be a Hardcoded credentials problem:https://securego.io/docs/rules/g101.html

We add this comment to tell gosec this is a false positive alarm when it scanning the code.

PR #358, opened 2023-06-22
Add the missing patch permission on clustermanagers

mprahl:


qiujian16:
please try the ocm repo instead, since all the code is moved to ocm repo now.

mprahl:
@qiujian16 thanks. I missed that it moved in the read me.

PR #363, opened 2024-09-20
Fix error message

jnpacker:
Small fix to the error message.


qiujian16:
@jnpacker we will need to put fix in the https://github.com/open-cluster-management-io/ocm/blob/main/pkg/operator/operators/crdmanager/manager.go

jnpacker:
The work has moved over here: https://github.com/open-cluster-management-io/ocm/pull/628

